PR #1, opened 2019-02-25
Adds reference to AUR package for Arch Linux

monsieurh:


README.md
```diff
@@ -2,11 +2,26 @@
...
 meson build --prefix=/usr --buildtype=release
 ninja -C build
 ```
-Now you can just run `./build/wf-recorder` or install it with `sudo ninja -C build install`. Use `-f <filename>` to specify the output file. In case of multiple outputs, you'll first be prompted to select the output you want to record. If you know the output name beforehand, you can use the `-o <output name>` option. 
+Now you can just run `./build/wf-recorder` or install it with `sudo ninja -C build install`.
+
+# usage
+Use `-f <filename>` to specify the output file. In case of multiple outputs, you'll first be prompted to select the output you want to record. If you know the output name beforehand, you can use the `-o <output name>` option. 
+
+You can use `wf-recorder -g (slurp)` to select and limit the recording to a part of the screen.
```
emersion: Should be `wf-recorder -g $(slurp)`

ammen99: actually I think it should be `wf-recorder -g "$(slurp)"` at least that's what I use for zsh

ammen99: also I think you should note that this requires `https://github.com/emersion/slurp`

monsieurh: You're right, I use fish and it happens to me all the time. My bad.

ammen99:
Thanks!

PR #4, opened 2019-03-10
Use sigemptyset() to initialize sigset_t

t6:
sigisemptyset() is a glibc only function per [1] that can be used
to check if a sigset_t is empty which is not needed here.  This
also fixes the build on FreeBSD where sigisemptyset() is not
available.

[1] https://linux.die.net/man/3/sigemptyset


ammen99:
Thanks!

PR #5, opened 2019-03-12
src: untabify

buhman:
The source previously mixed tabs and spaces. This normalized the source to
spaces. Tabs are fine too, but not both!


buhman:
Calculated that -> spaces would be the smaller diff via eyeballs; unconfirmed.

buhman:
@ammen99 thanks!

Issue #2, opened 2019-02-26
Specifying a video codec


hedgepigdaniel:
Thanks for making this btw :)

It would be great if it was possible to choose the ffmpeg codec. The reason I want that is that I have an intel GPU and I'd like to record the screen with low/negligible extra CPU usage by using the GPU's video encoding ability. This is possible with ffmpeg: https://trac.ffmpeg.org/wiki/Hardware/QuickSync

ammen99:
Is changing the encoder enough? For example if you set it manually here:
https://github.com/ammen99/wf-recorder/blob/master/src/frame-writer.cpp#L48

hedgepigdaniel:
I remember when using https://github.com/atomnuker/wlstream that changing the video codec usually resulted in having to change other options like pixel formats and the like. But essentially yes, replacing `libx264` with `h264_qsv` should make it work.

Potentially I'll try changing the hard coded values and see if I can find a set of options that works.

foxcpp:
It would be much more user-friendly to be able to set codec options from command line.

ammen99:
@foxcpp I am not very knowledgeable about ffmpeg, but if somebody could point me what codecs/options we need, I will gladly add support for command line arguments.

foxpy:
> but if somebody could point me what codecs/options we need

We **totally need** `libx264` as the most portable option which provides good balance of performance, quality and file size. It's included already, fine! (though it would be nice to let user change crf and preset values).
We _may need_ `libx265` for people with multi-core horsepower (like high end Intel Core or Ryzen CPU) to fit more image quality into smaller file size. Also good thing that `libx265` has the same interface options like `libx264`.
We _may need_ `libvpx` (both VP8 and VP9) as free and open source alternative to closed H.264/H.265 format.

These were the software codecs which work everywhere. Now the GPU part...
The worst thing about GPU encoding is that it's not going to perform fine on every system there is. Also, API is going to be different for every platform.
ffmpeg supports VAAPI, VDPAU and NVENC, which are going to be suffient for screen recording in Wayland. VAAPI for Intel/AMD GPUs, VDPAU for NVIDIA/AMD GPUs and NVENC exclusively for NVIDIA GPUs.
Also I'm not even sure if VAAPI/VDPAU work in Wayland...

hedgepigdaniel:
[...]

PR #7, opened 2019-03-14
Add command line argument to specify the codec

hedgepigdaniel:
An attempt to address https://github.com/ammen99/wf-recorder/issues/2

Usage: `wf-recorder -c libx264` (default) or other ffmpeg codecs.

It works for me with the nvenc codecs, e.g. `wf-recorder -c nvenc_h264` or `wf-recorder -v nvenc_hevc`. In my case I have to do `optirun wf-recorder`, since I have nvidia optimus graphics and bumblebee

### Issues
- So far it does not work with the VAAPI encoders for Intel GPUs (which are more common, less power hungry, and don't require closed source drivers). I think this is because they require an ffmpeg hardware device to be initialised, and maybe also a different pixel format and filters to upload the frames to the device. See https://trac.ffmpeg.org/wiki/Hardware/VAAPI for command line examples. It works with wlstream: https://github.com/atomnuker/wlstream/blob/master/src/wlstream.c
- With nvenc, one of my CPU cores is maxed out when capturing
- the `crf` option is not valid for nvenc or vaapi encoders - so a different method is needed to control bitrate/quality

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -22,7 +22,7 @@ public :
 
 static FFmpegInitialize ffmpegInitialize;
 
-FrameWriter::FrameWriter(const string& filename, int width_, int height_, InputFormat format) :
+FrameWriter::FrameWriter(const string& filename, const std::string& codec_name, int width_, int height_, InputFormat format) :
     width(width_), height(height_)
 {
     switch (format)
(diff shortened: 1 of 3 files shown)


ammen99:
Thanks for looking into this, I didn't expect it to be "so easy".

Sadly `wlstream` might not be very helpful for us, because it uses the `export-dmabuf` approach which is zero-copy (with `wf-recorder`/screencopy protocol, the data is copied once to the CPU). And I used the screencopy protocol exactly because I didn't want to get involved in choosing a pixel format, codecs, etc. that work on most setups.

However I think we should still give an attempt at doing the encoding in HW for VAAPI. I might be able to find how to initialize a render node with ffmpeg, in which case we should be able to add yet another argument pointing to the device to use.

ammen99:
Another thought: the CPU usage probably comes from the fact I that the RGB data we get from the compositor is first converted to YUV.

ammen99:
Looking into it more, it seems like VAAPI support will not be as simple as just initializing a HW context. We also need to change the codec input format (seems neither YUV420p nor BGR0 are supported) and potentially other things.

ammen99:
I was able to get VAAPI support working, however I need to clean up the code, handle errors, etc. Thanks for this patch!

PR #11, opened 2019-03-17
Set sw_format in the hardware device context to NV12

hedgepigdaniel:


ammen99:
Thanks for finding this.

PR #15, opened 2019-03-30
Clean up

soreau:
Minor changes to fix warnings.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -616,8 +616,6 @@ int main(int argc, char *argv[])
 
     signal(SIGINT, handle_sigint);
 
-    std::cout << "start at " << timespec_to_usec(get_ct()) / 1.0e6<< std::endl;
-
     while(!exit_main_loop)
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -39,6 +39,14 @@ struct wf_recorder_output
 
 std::vector<wf_recorder_output> available_outputs;
 
+static timespec get_ct()
```
ammen99: is this function used at all?

soreau: It's used to output some info but that line could be removed with the entire function.
`std::cout << "start at " << timespec_to_usec(get_ct()) / 1.0e6<< std::endl;`

ammen99: yeah that was a debug print, we don't need it

soreau: Ok fixed, I just merged everything into a single commit.

ammen99:
Looks good, thanks!

Issue #6, opened 2019-03-14
Detect output when using slurp 


zajca:
Don't know if this is possible but would be great to detect output when slurp is used.

Or even record multiple output at once, for example grim can make screenshot from area including multiple outputs when using slurp.

ammen99:
Recording multiple outputs is possible, yes, but I am not sure it will work very well. First of all, different outputs can have different refresh rates. Frames might come at different times even if the refresh rate is the same. Plus, we'll need to manually compose the buffers together to present a single big buffer to ffmpeg. So except if I'm wrong I don't think multi-output-recording will be very useful.

About detecting which output, yes, this makes sense. I'll see what I can do.

ammen99:
This was fixed in #19. Reopen if something doesn't work.

PR #19, opened 2019-04-21
Add automatic output detection based on geometry

CrazedProgrammer:
The only thing that bugged me about wf-recorder was that it prompts the user to select an output even when having passed selection geometry. This broke my screen recording shortcut on my desktop where I have multiple monitors.  
This PR adds automatic output detection when the user passes geometry using `-g`.

Related to #6

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -453,7 +453,7 @@ struct capture_region
         return width > 0 && height > 0;
     }
 
-    bool contained_in(const capture_region& output)
+    bool contained_in(const capture_region& output) const
     {
         return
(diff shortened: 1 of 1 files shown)


ammen99:
Looks good, except please use `std::cout/std::cerr` instead of `fprintf`. I copied some of the code from a C program, that's why there are still `printf`s around, but new code should be with `cout`.

CrazedProgrammer:
Done! Thanks for the quick response :)

ammen99:
Thanks! This fixes #6

Issue #24, opened 2019-04-29
Way to start and stop recordings using keybindings


samerickson:
Hi there, I am using `swaywm` and as far as I can tell the way you stop the recording is by pressing `ctrl+c` but is there a way to start and stop the recording without the requirement of having a terminal open?

ammen99:
yes, you can just kill the process with `SIGINT` (which is exactly what ctrl-c does)

mathroc:
here is something that works for me with sway:

```
bindsym Ctrl+Shift+Print exec wf-recorder -g "$$(~/bin/slurp)" -f ~/recording_$(date +"%Y-%m-%d_%H:%M:%S.mp4")
bindsym Ctrl+Shift+BackSpace exec killall -s SIGINT wf-recorder
```

samerickson:
Thanks for both of your help, I have been playing around with it. @mathroc thanks for your snippet including `slurp`.

schauveau:
The timeout command, assuming that it is installed on your system, can also be useful to record for a given amount of time (or to limit the recording time to a reasonable value). For instance, to record during 10 minute: 
`timeout -s SIGINT 10m wf-recorder `

S1ngl3-x:
Late to the party but in case anyone else is looking a for **toggle** keybinding and uses fish scripts as I do then this works for me

` bindsym $mod+Shift+R exec --no-startup-id fish -c "pgrep -x wf-recorder > /dev/null; and pkill -SIGINT wf-recorder; or wf-recorder -g (slurp) -f ~/Videos/Recordings/screencast-(date +%Y-%m-%d_%H-%M-%S).mkv"`

PR #31, opened 2019-05-22
usage with sway

mathroc:
see #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,3 +38,17 @@ To use gpu encoding, use a VAAPI codec (for ex. `h264_vaapi`) and specify a GPU 
 ```
 wf-recorder -f test-vaapi.mkv -c h264_vaapi -d /dev/dri/renderD128
 ```
+
+## with [Sway](https://swaywm.org/)
+
+In your Sway config file, add those key bindings to start and stop wf-recorder :
+
+```
+bindsym Ctrl+Print exec wf-recorder -f ~/recording_$(date +"%Y-%m-%d_%H:%M:%S.mp4")
+bindsym Ctrl+Shift+Print exec wf-recorder -g "$$(slurp)" -f ~/recording_$(date +"%Y-%m-%d_%H:%M:%S.mp4")
+bindsym Ctrl+Shift+BackSpace exec killall -s SIGINT wf-recorder
+```
+
+* `Ctrl+Print` will start recording the whole screen
+* `Ctrl+Shift+Print` will let you select an area to record and start recording
+* `Ctrl+Shift+BackSpace` will stop all recordings


README.md
```diff
@@ -38,3 +38,17 @@ To use gpu encoding, use a VAAPI codec (for ex. `h264_vaapi`) and specify a GPU
...
 wf-recorder -f test-vaapi.mkv -c h264_vaapi -d /dev/dri/renderD128
 ```
+
+## with [Sway](https://swaywm.org/)
+
+In your Sway config file, add those key bindings to start and stop wf-recorder :
+
+```
+bindsym Ctrl+Print exec wf-recorder -f ~/recording_$(date +"%Y-%m-%d_%H:%M:%S.mp4")
+bindsym Ctrl+Shift+Print exec wf-recorder -g "$$(slurp)" -f ~/recording_$(date +"%Y-%m-%d_%H:%M:%S.mp4")
```
soreau: Is this double $$ a typo in `-g "$$(slurp)"`?

soreau:
This information might be a better fit for the wf-recorder wiki, just a thought.

README.md
mathroc: no it's for escaping the `$`, otherwise it's considered a configuration variable

mathroc:
I don't mind having this in the wiki but I don't think I can edit it. I'll let anyone interested either merging this PR or copying its content in the wiki

ammen99:
I agree we'd better add this to the Wiki page, since in the future we'll want to show integrations with more environments. I'm not sure if people can send PRs to the wiki.

norcalli:
I have a different version which allows you to use a single `bindsym` and a flock to be able to start and stop the video which I keep in a script. Would sharing such a script be a better solution? That way people can integrate with it however they want.

soreau:
@norcalli Can you upload it as a [gist](https://gist.github.com/) and post the link here?

ammen99:
I think we should really have a wiki page where this all belongs to (what this PR has, also what @norcalli has).

Just paste it to gist, and I'll upload it to the wiki (since AFAIK there are no pull requests for the wiki)

norcalli:
@soreau https://gist.github.com/norcalli/81fa49055767d84986637151237f7dbd the recording path is hardcoded to my preferences and uses slurp for region recording. Canceling slurp records the whole screen. Running the script again stops screen recording.

soreau:
> I think we should really have a wiki page where this all belongs to (what this PR has, also what @norcalli has).
> 
> Just paste it to gist, and I'll upload it to the wiki (since AFAIK there are no pull requests for the wiki)

I created a [wiki page](https://github.com/ammen99/wf-recorder/wiki/Usage-with-Sway) with this information.

ammen99:
@soreau Thank you. In this case we can close this PR.
Also thanks @mathroc for giving those tips, I think users will find them useful :)

PR #32, opened 2019-05-22
Add OpenCL support

soreau:
This attempts to use OpenCL to convert rgb data to yuv instead of sws_scale() when wf-recorder is built with OpenCL support and the --to-yuv or -t option is passed. This helps offload work to the gpu that would otherwise be done on the cpu.

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -385,10 +385,9 @@ void FrameWriter::add_frame(const uint8_t* pixels, int64_t usec, bool y_invert)
     if (hw_device_context)
     {
 #ifdef HAVE_OPENCL
-        uint32_t *local_yuv_buffer = NULL;
         if (params.to_yuv)
         {
-            int ret = opencl->do_frame(pixels, &local_yuv_buffer, encoder_frame, get_input_format(), y_invert);
+            int ret = opencl->do_frame(pixels, encoder_frame, get_input_format(), y_invert);
 
             if (ret)
                 sws_scale(swsCtx, &formatted_pixels, stride, 0, params.height,
@@ -409,10 +408,6 @@ void FrameWriter::add_frame(const uint8_t* pixels, int64_t usec, bool y_invert)
         }
 
         output_frame = &hw_frame;
-
-#ifdef HAVE_OPENCL
-	free(local_yuv_buffer);
-#endif
     } else if(get_input_format() == videoCodecCtx->pix_fmt)
     {
         output_frame = &encoder_frame;
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -123,7 +123,7 @@ AVPixelFormat FrameWriter::choose_sw_format(AVCodec *codec)
 {
 #ifdef HAVE_OPENCL
     /* First case: the user preference is to convert the rgb data to nv12 */
-    if (params.to_nv12 && is_fmt_supported(AV_PIX_FMT_NV12, codec->pix_fmts))
+    if (params.convert_rgb && is_fmt_supported(AV_PIX_FMT_NV12, codec->pix_fmts))
         return AV_PIX_FMT_NV12;
 #endif
     /* Second case: if the codec supports getting the appropriate RGB format
@@ -165,7 +165,7 @@ void FrameWriter::init_video_stream()
     videoCodecCtx->time_base = (AVRational){ 1, FPS };
 
 #ifdef HAVE_OPENCL
-     if (params.to_nv12)
+     if (params.convert_rgb)
          opencl = std::unique_ptr<OpenCL> (new OpenCL(params.width, params.height));
 #endif
 
@@ -175,7 +175,7 @@ void FrameWriter::init_video_stream()
         init_hw_accel();
         videoCodecCtx->hw_frames_ctx = av_buffer_ref(hw_frame_context);
 
-        if (params.to_nv12)
+        if (params.convert_rgb)
             init_sws(AV_PIX_FMT_NV12);
     } else
     {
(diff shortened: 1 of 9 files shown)


src/frame-writer.cpp
```diff
@@ -10,6 +10,14 @@
 #include <cstring>
 #include "averr.h"
 
+#include "config.h"
+
+#ifdef HAVE_OPENCL
+#include "opencl.hpp"
+
+std::unique_ptr<OpenCL> opencl;
```
ammen99: Can we not move this inside the `FrameWriter` class? Of course, this would also mean adding `#ifdef HAVE_OPENCL` checks to the `opencl.hpp` too

ammen99: You can probably even make the class empty when OpenCL is disabled (no members, just emtpy constructor) and then actually reduce the checks (for ex. you could remove the checks from meson in that case)

src/opencl.cpp
```diff
@@ -0,0 +1,242 @@
...
+ * Copyright 2019 Scott Moreau
+ *
+ */
+
+#include <iostream>
+
+#include "opencl.hpp"
+
+
+static char const *cl_source_str = "									\n\
```
ammen99: Instead of adding all of these `\n`, use C++11 raw strings.

src/frame-writer.cpp
soreau: I guess it's possible but for what reason? It just seems it would complicate the code.

src/opencl.cpp
soreau: I couldn't figure out how to do this. The call to clCreateProgramWithSource() expects a const char**.

ammen99:
I think that the options now are a bit confusing TBH.

For example, we have a `--no-opencl` which works only with `--convert-rgb`, and we'd need `--convert-rgb` to enable OpenCL with libx264 (so GPU converting and CPU encoding), which is of course strange. What would you say about an option `--opencl` (which enables OpenCL if it can be used) and another, `--force-yuv` for the radeon vaapi usecase?

soreau:
> I think that the options now are a bit confusing TBH.
>
>For example, we have a --no-opencl which works only with --convert-rgb, and we'd need 
>--convert-rgb to enable OpenCL with libx264 (so GPU converting and CPU encoding), which is of >course strange. What would you say about an option --opencl (which enables OpenCL if it can be >used) and another, --force-yuv for the radeon vaapi usecase?

I guess this is fine but in the case where opencl isn't built in, should we ifdef the option out? In the case where you'd use --force-yuv, I'm guessing you'd probably want to use opencl if it is built in, so you'd need to pass both options. You'd only need to disable opencl to test performance gains or if there's a problem.

I would like to have opencl enabled by default if it was built in and the other option could be called whatever but I would like --to-yuv.

src/opencl.cpp
soreau: I got it now, done.

soreau:
I made it like you suggest to see how it works out. I meant `-t` as --to-yuv and `-e` as --enable-opencl but have `--force-yuv` and `--opencl` for now. I also committed your patch to use a function to avoid code duplication and literal string for the opencl program.

src/frame-writer.cpp
```diff
@@ -355,6 +372,32 @@ FrameWriter::FrameWriter(const FrameWriterParams& _params) :
...
+    if (params.opencl && params.force_yuv)
+    {
+        int r = opencl->do_frame(pixels, encoder_frame,
+            get_input_format(), y_invert);
+
+        converted_with_opencl = (r == 0);
+    }
+#else
+    /* Silence compiler warning when opencl is disabled */
+    (void)(y_invert);
```
ammen99: we can move this inside the #ifdef

src/main.cpp
```diff
@@ -506,14 +507,15 @@ int main(int argc, char *argv[])
...
+        { "force-yuv",       no_argument,       NULL, 't' },
+        { "opencl",          no_argument,       NULL, 'e' },
         { 0,                 0,                 NULL,  0  }
     };
 
     int c, i;
     std::string param;
     size_t pos;
-    while((c = getopt_long(argc, argv, "o:f:g:c:p:d:la::t::", opts, &i)) != -1)
+    while((c = getopt_long(argc, argv, "o:f:g:c:p:d:la::t::e::", opts, &i)) != -1)
```
ammen99: why does `e` have optional parameter when it doesn't take any value?

soreau: Fixed.

src/frame-writer.cpp
soreau: The point of this is to not leave y_invert variable declared but not used in the case where opencl is disabled. If we put it inside the ifdef, the warning still happens.

ammen99:
I'll test this later today and hopefully merge, if there is nothing else that you have planned to do.

soreau:
I don't have anything else planned at this time.

soreau:
@ammen99 Can you test [this patch](http://ix.io/1KS2) (on top of opencl branch)? Adjust the value of SIZE_MULTIPLIER and see if there is a relation to how much cpu is being used. I've tried between 1-32, with and without vaapi using libx264 and h264_vaapi. Here the results are that the higher the number (meaning the more memory allocated and copied), the less cpu that is being consumed (via htop). It seems counterintuitive that copying more data would result in less cpu usage. Is it the case for you too? Any idea what might be going on here?

soreau:
As discussed in irc, it has been realized that the cpu was probably doing less because the gpu was doing more copying, essentially the same as calling usleep after copying. Now that I've set the buffers to the correct size, I've tested on a box with libx264 and h264_vaapi, with and without opencl, while idle and while moving mouse cursor, recording the top two (highest cpu usage) threads and the results are always better in the opencl case except libx264+sws_scale when idle. It also should now clean up properly without crashing.
```
                    |   sws_scale  |    OpenCL    |
                    |______________|______________|
                    |   T1   T2    |   T1   T2    |
____________________|______________|______________|
           | idle   |   95%  50%   |   97%  54%   |
libx264    |        |              |              |
           | motion |   57%  32%   |   51%  26%   |
___________|________|______________|______________|
           |        |              |              |
           | idle   |   47%  46%   |   38%  37%   |
h264_vaapi |        |              |              |
           | motion |   38%  38%   |   32%  32%   |
```

ammen99:
So this should be ready for merging, right? Could you please rebase on top of master?

soreau:
> So this should be ready for merging, right? Could you please rebase on top of master?

Yes, should be ready to go now.

ammen99:
I tested this final version with both OpenCL on and off, and both VAAPI on and off, seems to work fine.

Thanks for the work on this, and sorry for the delay.

soreau:
> I tested this final version with both OpenCL on and off, and both VAAPI on and off, seems to work fine.
> 
> Thanks for the work on this, and sorry for the delay.

No problem, thanks for testing. I'd like to see this 'fixed' in the gallium vaapi driver and test the internal converter for comparison, but maybe that will come in the future.

Issue #26, opened 2019-05-12
Help / man page


clarfon:
Would be nice if wf-recorder at least offered a -h/--help command-line option, as it seems that it doesn't do that by default. Would also be nice to have a man page as well with the docs from the README.

Jpfonseca:
@ammen99  made you can add a enhancement label on this one

ammen99:
Fixed in #35

PR #30, opened 2019-05-21
Add to-yuv option

soreau:
Add a new option --to-yuv (or -t) which converts each frame from rgb0/bgr0 to nv12/yuv420
in software using sws_scale(), before sending to the gpu for encoding. This is needed for
some drivers like the radeonsi vaapi driver, which crashes when fed rgb0 data for encoding.
See fdo bug #110719. Fixes #13.

PR #35, opened 2019-06-12
wf-recorder --help / manpage

Jpfonseca:
Started to contribute to fulfill issue #26

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -550,8 +550,10 @@ int main(int argc, char *argv[])
             case 't':
                 params.to_yuv = true;
                 break;
+            
             case 'h':
                 break;
+                
             case 'p':
                 param = optarg;
                 pos = param.find("=");
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -481,6 +481,68 @@ static wf_recorder_output* detect_output_from_region(const capture_region& regio
     return nullptr;
 }
 
+void help(){
+  
+    printf("Usage: wf-recorder [OPTION]... -f [FILE]...
+Screen recording of wlroots-based compositors
+
+With no FILE, start recording the current screen.
+
+  -a, --audio [OPTION]      Starts recording the screen with audio.
+                            [OPTION] In case you want to specify the pulseaudio device which will capture the audio, 
+                            you can run this command with the name of that device. 
+                            You can find your device name by running : pactl list sinks | grep Name
+                
+  -c, --codec               Specifies the codec of the video. Supports  GIF output also.
+                            To modify codec parameters, use -p <option_name>=<option_value>
+                
+  -d, --device              Selects the device to use when encoding the video
+                            Some drivers report support for rgb0 data for vaapi input but really only support yuv.
+                            Use the -t or --to-yuv option in addition to the vaapi options to convert the data in software,
+                            before sending it to the gpu.
+ 
+  -f <filename>.ext         By using the -f option the output file will have the name filename.ext and the file format will be 
+                            determined by provided file extension .ext . If the extension .ext provided is not recognized by your FFmpeg muxers,
+                            the command will fail.
+                            You can check the muxers that your FFmpeg installation supports by running  : ffmpeg -muxers  
+           
+  -g, --geometry            Selects a specific part of the screen.
+  
+  -h, --help                Prints this help screen.
+                
+  -l, --log                 Generates a log on the current terminal. Debug purposes. 
+                
+  -o, --output              Specify the output where the video is to be recorded.
+                
+  -p, --codec-param         Change the codec parameters.
+                            -p <option_name>=<option_value>
+
+  -t, to-yuv                Use the -t or --to-yuv option in addition to the vaapi options to convert the data in software,
+                            before sending it to the gpu.
+                
+                
+                
+Examples:
+  
+  Video Only:
+  
+  - wf-recorder                        Records the video. Use Ctrl+C to stop recording. 
+                                       The video file will be stored as recording.mp4 in the current working directory.
+                
+  - wf-recorder -f <filename>.ext      Records the video. Use Ctrl+C to stop recording. 
+                                       The video file will be stored as <outputname>.ext in the current working directory.
+ 
+  Video and Audio:
+  
+  - wf-recorder -a                     Records the audio. Use Ctrl+C to stop recording. 
+                                       The video file will be stored as recording.mp4 in the current working directory.
+                
+  - wf-recorder -a -f <filename>.ext   Records the audio. Use Ctrl+C to stop recording. 
+                                       The video file will be stored as <outputname>.ext in the current working directory.
+");
+        exit(EXIT_SUCCESS);
+}
+
 int main(int argc, char *argv[])
 {
     FrameWriterParams params;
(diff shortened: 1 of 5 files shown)


src/main.cpp
```diff
@@ -507,13 +507,14 @@ int main(int argc, char *argv[])
...
         { "to-yuv",          no_argument,       NULL, 't' },
+        { "help",            no_argument,       NULL, 'h' },
         { 0,                 0,                 NULL,  0  }
     };
 
     int c, i;
     std::string param;
     size_t pos;
-    while((c = getopt_long(argc, argv, "o:f:g:c:p:d:la::t::", opts, &i)) != -1)
+    while((c = getopt_long(argc, argv, "o:f:g:c:p:d:la::t::h::", opts, &i)) != -1)
```
soreau: From the getopt_long man page `Two colons mean an option takes an optional arg`. This option doesn't take an optional argument, does it?

src/main.cpp
```diff
@@ -563,12 +565,65 @@ int main(int argc, char *argv[])
                     printf("Invalid codec option %s\n", optarg);
                 }
                 break;
-
+
```
soreau: Don't add random whitespace.

src/main.cpp
```diff
@@ -563,12 +565,65 @@ int main(int argc, char *argv[])
...
                 break;
-
+            
             default:
                 printf("Unsupported command line argument %s\n", optarg);
         }
     }
-
+    
+    if(c=='h'){
```
soreau: Use spaces around `==` and after `)`.

src/main.cpp
```diff
@@ -563,12 +565,65 @@ int main(int argc, char *argv[])
...
+         printf("Usage: wf-recorder [OPTION]... [FILE]...
```
soreau: This printf statement is not aligned with the rest of the code.

src/main.cpp
```diff
@@ -563,12 +565,65 @@ int main(int argc, char *argv[])
...
+  -g, --geometry            Selects a specific part of the screen.
+  
+  -h, --help                Prints this help screen.
+                
+  -l, --log                 Generates a log on the current terminal. Debug purposes. 
+                
+  -o, --output              Specify the output where the video is to be recorded.
+                
+  -p, --codec-param         Change the codec parameters.
+                       -p <option_name>=<option_value>
```
soreau: The alignment is incorrect here too. Use spaces instead of tabs.

src/main.cpp
Jpfonseca: Sure misclick on that. Btw can I change the options to be in alphabetical order ?

Jpfonseca:
Made the changes needed. Can I change the options to be in alphabetical order ?

src/main.cpp
```diff
@@ -549,7 +550,10 @@ int main(int argc, char *argv[])
             case 't':
                 params.to_yuv = true;
                 break;
-
+
```
soreau: This line adds whitespace, several spaces on a blank line.

src/main.cpp
```diff
@@ -569,6 +573,59 @@ int main(int argc, char *argv[])
         }
     }
 
+    if ( c == 'h' ) {
```
soreau: Don't put space after `(` or before `)`.

soreau:
> Can I change the options to be in alphabetical order ?

I think this is a different idea that should be in a separate PR. @ammen99 might have different ideas than me, what I say is not absolute.

soreau:
> @soreau I just looked to the rest of the file and changed the next if to follow the same code style also

I notice the file is quite inconsistent in this regard, the brackets should all be the same style though I don't know which way is correct. Really need style guidelines somewhere. @ammen99?

ammen99:
I'm on a trip so I can't review properly atm, I'll be back in 3 days. Anyways, brackets should be on the next line.

Jpfonseca:
Will keep that in mind.

src/main.cpp
```diff
@@ -569,8 +573,63 @@ int main(int argc, char *argv[])
         }
     }
 
+    if (c == 'h')
```
ammen99: Please move this to a separate function.

src/main.cpp
```diff
@@ -569,8 +573,63 @@ int main(int argc, char *argv[])
         }
     }
 
+    if (c == 'h') 
+    {
+         printf("Usage: wf-recorder [OPTION]... [FILE]...
```
ammen99: `[FILE]` is passed with the `-f` parameter, not standalone.

src/main.cpp
```diff
@@ -569,8 +573,63 @@ int main(int argc, char *argv[])
...
+                
+                
+Examples:
+  
+  Video Only:
+  
+  - wf-recorder                       Records the video. Use Ctrl+C to stop recording. 
+                                      The video file will be stored as recording.mp4 in the current working directory.
+                
+  - wf-recorder -f <filename>         Records the video. Use Ctrl+C to stop recording.
```
ammen99: The output name can be another one. For example, the user can also set `wf-recorder -f test.mkv` and then the resulting file will be `test.mkv` not `test.mp4`

src/main.cpp
```diff
@@ -569,8 +573,63 @@ int main(int argc, char *argv[])
...
     }
 
+    if (c == 'h') 
+    {
+         printf("Usage: wf-recorder [OPTION]... [FILE]...
+Screen recording of wlroots-based compositors
+
+With no FILE, start recording the current screen.
+
+  -a, --audio               Starts recording the screen with audio
```
ammen99: For the sake of completeness, the `--audio/-a` parameter takes an optional parameter, which is the name of the pulseaudio device to capture (the list can be obtained by running `pactl list sinks | grep Name`)

src/main.cpp
```diff
@@ -569,8 +573,63 @@ int main(int argc, char *argv[])
...
+  -h, --help                Prints this help screen.
+                
+  -l, --log                 Generates a log on the current terminal. Debug purposes. 
+                
+  -o, --output              Specify the output where the video is to be recorded.
+                
+  -p, --codec-param         Change the codec parameters.
+                            -p <option_name>=<option_value>
+
+  -t, to-yuv                Use the -t or --to-yuv option in addition to the vaapi options to convert the data in software,
```
ammen99: Maybe leave this out from this PR until we merge #32, then me or @soreau will add the docs for `-t` and `-e`

src/main.cpp
Jpfonseca: On the first example no arguments where given. And for what I have experienced it always results in the test.mp4. Probably I can add the information of the inexistence of  arguments

ammen99: It is just that the default file name is `recording.mp4`. If you specify a `-f test.mkv` then it should produce `test.mkv`. If it doesn't, then this is a bug.

Jpfonseca: Let me clarify, I am giving 2 different examples on this print.

The first with the default value of the command `wf-recorder` without any arguments.
On the  the second, with a specific filename by running the command `wf-recorder -f <filename>` which will result in a recording with the name `filename`

soreau: Perhaps you could elaborate the comment for the -f case (or both comments if necessary) to explain that `-f file.ext` means that the output file will be called `file.ext` and the file format will be guessed by the file extension `.ext`, which will fail if it's not one listed in the output of `ffmpeg -muxers`.

ammen99: The whole point is that the output file is not necessarily a .mp4 file, as the current help message says(and as far as I understand it)

src/main.cpp
soreau: @ammen99 #30 was merged so -t already exists.

src/main.cpp
Jpfonseca: @ammen99  Sorry I only saw the .mp4 on the second example now, as I was answering on my phone :/ .  You were right  from the beginning. 
@soreau  I will try to address that.

src/main.cpp
Jpfonseca: Didn't add the -e option

Jpfonseca:
> I think it's safe to omit `Supports GIF output also.`. You could mention the list can be found with `ffmpeg -encoders`, though.

GIF support was the main reason why I started using this tool. If I hadn't read an issue on it, I wouldn't be using it

Jpfonseca:
@soreau  just tested everything and made the changes you asked

src/main.cpp
```diff
@@ -481,6 +481,74 @@ static wf_recorder_output* detect_output_from_region(const capture_region& regio
     return nullptr;
 }
 
+static void help(){
```
ammen99: new line :)

src/main.cpp
```diff
@@ -481,6 +481,74 @@ static wf_recorder_output* detect_output_from_region(const capture_region& regio
...
+
+    printf("\
```
ammen99: You can use C++11 raw strings here. Just do

```
printf(R"(
blah
blah
)");

```

In this case, you won't have to add `\n\` to every line.

Jpfonseca: Will change it .  Didn't knew about this improvement. Also was took too much time doing the `\n\ ` way...

Jpfonseca:
Applied all the changes needed. 
@ammen99 any ideas on what to use to make the man page ?  `scdoc` ?
That probably the best way to keep the manpage simple stupid whilst not messing much with the `meson.build` file .

ammen99:
I'd use whatever sway uses(which I think is scdoc)

Jpfonseca:
Added first version of the manpage.
@ammen99  @soreau do you think I should change anything else ? Or add something ?

Jpfonseca:
Was a good ride. At least I now know how to make a manpage 🎉.

PR #36, opened 2019-06-23
Changed the location of the manpage to avoid build error

Jpfonseca:
Build error, as the path was not correct.

ninja: error: '../wf-recorder/manpage/wf-recorder.1.scd', needed by 'wf-recorder.1', missing and no known rule to make it
==> ERROR: A failure occurred in build().
    Aborting...
Error making: wf-recorder-git


Jpfonseca:
error on the `meson.build` file

ammen99:
Thanks!

hughwilliams94:
Works for me!

PR #42, opened 2019-07-12
frame-writer: try to guess format for streaming

### LINKED ISSUE #40: non-file output for streaming purposes
I'm actually using the fork from @schauveau but I started with this branch, and he doesn't have issues enabled on his repository.  :-P

I tried to hack in support for non-file output, but it looks like one cannot mux without seeking.  I'd be satisfied with a pipe, but again, without the ability to seek...  Any tips on how to go about adding this functionality?  I'm shooting in the dark when it comes to the ffmpeg API.

/cc @norcalli

ammen99:
Fix #40 

@disruptek Can you test this? You should be able to directly say the rtmp/udp address to stream to, no need for additional parameters.

EDIT: I have tested with twitch only, and it works fine.

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -300,6 +300,16 @@ void FrameWriter::init_sws(AVPixelFormat format)
     }
 }
 
+static const char* determine_output_format(const std::string& output_name)
+{
+    if (output_name.find("rtmp") == 0)
+        return "flv";
+    if (output_name.find("udp") == 0)
+        return "mpegts";
+
+    return NULL;
+}
+
 FrameWriter::FrameWriter(const FrameWriterParams& _params) :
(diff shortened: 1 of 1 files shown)


colemickens:
I'm building this with #43 but only testing this functionality, I think. Seems to be working for me, I was able to stream to Twitch for a while at least from Sway.

disruptek:
I haven't tested the output, but it appears to be streaming.  I've had intermediate (and silent) stream failure, so I need to find a way to identify when it fails before I can return to putting this through its paces.  Thanks for the patch, though.  :wink:

ammen99:
Okay so I'll merge this. Thanks for testing.

 @disruptek Reopen #40 if it doesn't work.

PR #47, opened 2019-08-24
Update to avcodec_send_frame API

soreau:


src/frame-writer.cpp
```diff
@@ -171,13 +171,14 @@ void FrameWriter::init_video_stream()
         videoCodecCtx->flags |= AV_CODEC_FLAG_GLOBAL_HEADER;
 
     int err;
+    char str[256];
     if ((err = avcodec_open2(videoCodecCtx, codec, &options)) < 0)
     {
-        std::cerr << "avcodec_open2 failed: " << averr(err) << std::endl;
+        av_strerror(err, str, 256);
```
ammen99: seems like our av_err does exactly this?

src/frame-writer.cpp
```diff
@@ -279,9 +280,13 @@ void FrameWriter::init_codecs()
     }
 
     AVDictionary *dummy = NULL;
-    if (avformat_write_header(fmtCtx, &dummy) != 0)
+    char err[256];
+    int ret = avformat_write_header(fmtCtx, &dummy);
+    if (ret != 0)
     {
         std::cerr << "Failed to write file header" << std::endl;
+        av_strerror(ret, err, 256);
```
ammen99: same aas above, `averr()` should give you a string of the error.

src/frame-writer.cpp
```diff
@@ -447,15 +476,11 @@ void FrameWriter::add_frame(const uint8_t* pixels, int64_t usec, bool y_invert)
     pkt.data = NULL;
     pkt.size = 0;
 
-    int got_output;
-    avcodec_encode_video2(videoCodecCtx, &pkt, *output_frame, &got_output);
+    encode(videoCodecCtx, *output_frame, &pkt);
 
     /* Restore frame buffer, so that it can be properly freed in the end */
     if (saved_buf0)
```
ammen99: IIRC this should happen before `finish_frame` as was in the original code, otherwise we may be leaking memory.

src/frame-writer.cpp
soreau: No, it does not work, averr gives garbage. For example, if I don't set LIBVA_DRIVER_NAME, I hit https://github.com/ammen99/wf-recorder/blob/master/src/frame-writer.cpp#L36 and the terminal shows "Failed to create hw encoding device : Input/ou�w}��y". If I use av_strrerror(), I get this: "Failed to create hw encoding device : Input/output error"

src/frame-writer.cpp
soreau: Same as above, averr gives garbage results.

src/frame-writer.cpp
soreau: `finish_frame` is called in encode() in a while loop, after avcodec_send_frame(). I think if you look how to use avcodec_send_frame() API, you will see this is correct.

src/frame-writer.cpp
soreau: Maybe it has something to do with [this comment regarding av_err2str()](https://www.ffmpeg.org/doxygen/4.1/group__lavu__error.html#gac4f6e109c242c81aeee21868d3f35e12): "Convenience macro, the return value should be used only directly in function arguments but never stand-alone." Since you're wrapping it in a function and then using that function, might this be a problem?

src/frame-writer.cpp
```diff
@@ -392,6 +397,30 @@ void FrameWriter::convert_pixels_to_yuv(const uint8_t *pixels,
     }
 }
 
+void FrameWriter::encode(AVCodecContext *enc_ctx, AVFrame *frame, AVPacket *pkt)
+{
+    int ret;
+
+    /* send the frame to the encoder */
+    ret = avcodec_send_frame(enc_ctx, frame);
+    if (ret < 0) {
```
ammen99: style: { on new line

src/frame-writer.cpp
```diff
@@ -392,6 +397,30 @@ void FrameWriter::convert_pixels_to_yuv(const uint8_t *pixels,
...
+        fprintf(stderr, "error sending a frame for encoding\n");
+        return;
+    }
+
+    while (ret >= 0) {
```
ammen99: again new line before {

src/frame-writer.cpp
```diff
@@ -392,6 +397,30 @@ void FrameWriter::convert_pixels_to_yuv(const uint8_t *pixels,
...
+        ret = avcodec_receive_packet(enc_ctx, pkt);
+        if (ret == AVERROR(EAGAIN) || ret == AVERROR_EOF)
```
ammen99: missing braces around if

src/frame-writer.cpp
ammen99: Just looked into it again, seems you're right.

soreau:
> Shouldn't we also update the audio encoding to the same API?

Perhaps, though I am unsure of what might be different from using this API for video.

src/frame-writer.cpp
colemickens: `ret` is not defined?

colemickens:
This is a very slightly modified version of this PR that fixes compilation and seems to get ffmpeg+vaapi working with ffmpeg4, although it still uses a ton of my CPU...

https://github.com/colemickens/wf-recorder/tree/ffmpeg4-vaapi

src/frame-writer.cpp
soreau: @colemickens Nice catch, not sure what happened there. Fixed.

soreau:
> This is a very slightly modified version of this PR that fixes compilation and seems to get ffmpeg+vaapi working with ffmpeg4, although it still uses a ton of my CPU...
> 
> https://github.com/colemickens/wf-recorder/tree/ffmpeg4-vaapi

Can you say what else is changed besides the build fix?

colemickens:
@soreau it's just the tiny build fix and a rebase, you can diff our branches.

However, it seems to sort of stop working after sometime. I only tried a few times last night before switching gears, but when I used vaapi, the stream would work for a few minutes and then stop. The process would keep eating CPU, but Twitch would just be frozen on a static image.

soreau:
> @soreau it's just the tiny build fix and a rebase, you can diff our branches.

Ok.

> However, it seems to sort of stop working after sometime. I only tried a few times last night before switching gears, but when I used vaapi, the stream would work for a few minutes and then stop. The process would keep eating CPU, but Twitch would just be frozen on a static image.

See if you can check if it's eating a lot of memory. Maybe it has a memory leak somewhere.

soreau:
I have updated video and audio to the new API.

PR #50, opened 2019-09-17
Update README.md: fix typo in clone url

RMPR:


soreau:
The url should work with or without the .git suffix. It is not required.

ammen99:
Well I don't see any harm, although I don't see why this matters either. 

@RMPR can you elaborate why we need this?

RMPR:
I tried without the .git and had a timeout, it may be my Internet connection though, but integrating those changes is not harmful since it's formatted like that inside the clone dialog.

2019-09-02-100204_grim

RMPR:
When I tried without the .git I had a timeout (it may be my Internet connection though) but it's not harmful since it's the way github formats it inside the clone dialog.
2019-09-17-191932_grim

ammen99:
I'm pretty sure this must have been a network issue on your side, but you have a point about how github formats this by default. Thanks for the PR!

PR #54, opened 2019-10-04
Opencl fixup

soreau:


src/frame-writer.cpp
```diff
@@ -170,6 +170,9 @@ void FrameWriter::init_video_stream()
     videoCodecCtx->height = params.height;
     videoCodecCtx->time_base = (AVRational){ 1, FPS };
 
+    if (params.force_yuv || params.opencl)
```
ammen99: I would like to see a comment explaining why this is necessary (or what it is a workaround for)

src/main.cpp
```diff
@@ -561,6 +571,7 @@ With no FILE, start recording the current screen.
                                         The video file will be stored as <outputname>.ext in the
                                         current working directory.)"
     );
+    printf("\n");
```
ammen99: Just merged #52 I think this shouldn't be necessary anymore.

src/main.cpp
```diff
@@ -535,10 +535,20 @@ With no FILE, start recording the current screen.
...
 
   -p, --codec-param         Change the codec parameters.
-                            -p <option_name>=<option_value>
+                            -p <option_name>=<option_value>)");
+#ifdef HAVE_OPENCL
+    printf(R"(
+
+  -e, --opencl              Use the -e[#] or --opencl[=#] in conjunction with -t or --force-yuv option
+                            to use opencl for gpu accelerated conversion of data to yuv. # is one
+                            of the devices listed when running without specifying #.)");
```
ammen99: Have you tested this? I think you would need a newline at the end of this.

src/main.cpp
```diff
@@ -535,10 +535,20 @@ With no FILE, start recording the current screen.
...
+#endif
+    printf(R"(
 
-  -t, force-yuv             Use the -t or --force-yuv option to force conversion of the data to
-                            yuv format, before sending it to the gpu.\n\n
+  -t, --force-yuv           Use the -t or --force-yuv option to force conversion of the data to
+                            yuv format, before sending it to the gpu.)");
+    printf("\n\n");
```
ammen99: Instead of printing \n\n why not just add two lines in the source and make it one whole printf?

soreau: Unless there's a way to escape \n character, it will print \n literally.

src/frame-writer.cpp
soreau: This is for drivers like radeonsi that do not support b-frames. Without this, attempting to use opencl will throw an error message that b-frames are not supported. I'm thinking we might istead add a `-b` option the user can pass that will be fed to `max_b_frames` and just not set it at all if `-b` is not passed. What do you think?

src/main.cpp
soreau: It was working when I tested it, yes.

src/main.cpp
ammen99: Just add more lines

soreau:
> Should also add `-b` to the help page.

Indeed, done.

ammen99:
Thanks a lot!

Issue #55, opened 2019-10-11
wf-recorder might end up trying to encode odd widths/heights if display scaling isn't an integer


mortie:
I have a 4k screen with 1.5x scaling, because stuff is too small at 1x and too big at 2x. Since wf-recorder deals with regions in logical units, even though it makes sure the logical region has an even width and height, that width/height times 1.5 might end up odd. For example, if I provide a 663x590 region, wf-recorder will change that to 664x590, and then the region that's actually captured will be (664\*1.5)x(590\*1.5) which is 996x885.

I think the best way to fix this would be to translate from logical units to physical units as soon as possible, and then do all modifications on those physical units.

Eluminae:
Hello there!
I'd like to know if there is still works on this issue. Got the same project so I can't really use regioned records cause it will not work 1/2 times.
It is in pipes ? :angel:

ammen99:
@Eluminae, there is a pull request at #56 but I have requested changes since I think the approach there is a hack which is not guaranteed to work AFAICT.

mortie:
Yeah, I didn't take the time to look deeper into it, so I kind of just dropped it. The code generally works though, so cloning https://github.com/mortie/wf-recorder will give you a version of wf-recorder which (in practice) works consistently on HiDPI. It's a few commits behind at this point though.

Now, regarding whether it's guaranteed to work or not; your criticism was that the screencopy protocol says that 'The client should then create a buffer with the provided attributes, and send a "copy" request'. The attributes in question are format, width, height and stride. Looking at my code, that's actually what it does: it creates the `wl_buffer` with `buffer.wl_buffer = create_shm_buffer(format, width, height, stride, &buffer.data);`, where format, width, height and stride are just the values we got straight from the wlr_screencopy frame event. It's only the values in the `wf_buffer` instances which are sometimes modified. I think my approach is OK according to the screencopy protocol.

I should've written this back in october, but, well, work and life got in the way and I forgot about it.

PR #56, opened 2019-10-17
Make sure physical units, not logical units, are even

mortie:
The current wf-recorder just makes sure that the logical geometry provided by the user are even, but that doesn't guarantee that the dimensions of the buffer provided by zwlr_screencopy are even. With this patch, wf-recorder instead makes sure that the width and height of the physical buffer received from zwlr_screencopy is even.

This fixes https://github.com/ammen99/wf-recorder/issues/55.

There are a couple of behavior changes:

* The old code would potentially increase the width/height by one logical pixel. The new code potentially decreases the width/height by one physical pixel instead.
* The new code doesn't print `Adjusted geometry: %d,%d %dx%d`. If it's required, I could add it back, but I haven't because that would require adding global state to avoid printing the message for every new frame.

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -422,7 +422,7 @@ void FrameWriter::convert_pixels_to_yuv(const uint8_t *pixels,
 void FrameWriter::add_frame(const uint8_t* pixels, int64_t usec, bool y_invert)
 {
     /* Calculate data after y-inversion */
-    int stride[] = {int(4 * params.width)};
+    int stride[] = {int(params.stride)};
     const uint8_t *formatted_pixels = pixels;
     if (y_invert)
     {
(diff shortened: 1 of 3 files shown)


mortie:
Yeah, you're right. That doesn't make sense.

I also found out that my patches didn't work, which is strange because I could've sworn I tested it; maybe I just checked if recording videos worked or not without checking the video after to see if it was correct. It turns out that the FrameWriter just assumes that `stride = width * 4`, which obviously isn't correct when I changed the width like I did.

I've now removed the stride modification, and adjusted FrameWriter to take a stride parameter in addition to width/height.

mortie:
There, rebased :)

WhyNotHugo:
Thanks!

PR #57, opened 2019-10-17
Add m (muxer) and x (pixel-format) to getopt_long

soreau:


ammen99:
For the record, these options were implemented in f81bb8d27 , but weren't added to getopt_long.

Thanks for the catch!

PR #82, opened 2020-04-08
Only update when there is damage

any1:
Doing this reduces baseline CPU load when not much is happening on the screen. I'm not sure how different video formats behave w.r.t. variable frame rate, so this might not be a good idea after all, but I'm sending this in for review in case it happens to be useful.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -194,7 +194,7 @@ static void frame_handle_buffer(void *, struct zwlr_screencopy_frame_v1 *frame, 
         exit(EXIT_FAILURE);
     }
 
-    zwlr_screencopy_frame_v1_copy(frame, buffer.wl_buffer);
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -215,11 +215,18 @@ static void frame_handle_failed(void *, struct zwlr_screencopy_frame_v1 *) {
     exit_main_loop = true;
 }
 
+static void frame_handle_damage(void *data,
+    struct zwlr_screencopy_frame_v1 *frame, uint32_t x, uint32_t y,
```
ammen99, line 219: Please omit the names of the variables, otherwise we get a bunch of warnings for unused parameters.

ammen99:
I just published a small release because with this change, we effectively require `wlroots >= 0.9` (which is fine, I think people using such software generally keep their system relatively up-to-date).

Thanks for the patch!

PR #83, opened 2020-04-15
Add Fedora dependency install instructions.

Azza292:
- Adds Fedora instructions for building from source.

- Spelling fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,12 +19,19 @@ aura -A wf-recorder-git
 
 ## From Source
 =======
-### install dependences (ubuntu)
+### Install Dependencies
+
+#### Ubuntu
 ```
 sudo apt install libavutil-dev libavcodec-dev libavformat-dev libswscale-dev libpulse-dev
 ```
 
-### download & build
+#### Fedora
+```
+$ sudo yum install wayland-devel wayland-protocols-devel ffmpeg-devel
+```
+
+### Download & Build
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,12 +19,19 @@ aura -A wf-recorder-git
...
+
+#### Ubuntu
 ```
 sudo apt install libavutil-dev libavcodec-dev libavformat-dev libswscale-dev libpulse-dev
 ```
 
-### download & build
+#### Fedora
+```
+$ sudo yum install wayland-devel wayland-protocols-devel ffmpeg-devel
```
ammen99, line 31: yum? Should be dnf.

Azza292:
```suggestion
$ sudo dnf install wayland-devel wayland-protocols-devel ffmpeg-devel
```
`yum` worked for me on Fedora 31. Your call

ammen99: yum has been deprecated since a while. I am not sure how long it will work, probably the Fedora guys will be supporting it as alias for dnf for a long time, but I see no reason to use yum.

ammen99:
Alright, looks good with dnf. Thanks for compiling the list of deps :)

Azza292:
Thanks for the info regarding `yum` 🙂

README.md
Azza292: Oh right, wasn't aware - Thanks for the info 🙂

PR #84, opened 2020-04-15
Add audio example to README

mb720:
Hi!

I think being able to also record audio is a huge feature of wf-recorder!

That's why I'd like to offer an example that creates a video *with audio* for the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,6 @@ sudo dnf install wf-recorder
 ```
 
 ## From Source
-=======
 ### Install Dependencies
 
 #### Ubuntu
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +44,12 @@ To select a specific part of the screen you can either use `-g <geometry>`, or u
 wf-recorder -g "$(slurp)"
 ```
 
+You can record screen and sound at the same with
+
+```
+wf-recorder --audio --file=$HOME/recording_$(date "+%F_%H-%M-%S").mp4
```
ammen99, line 50: I think this is a good idea but we should keep examples as simple as possible, so please drop the fancy file name. Maybe just `with_audio.mp4`

mb720: I admit the file name looks fancy^^. But it's also very useful to have the current date and time in it: no more overwriting of recordings. I also do this for my screenshots. Do you want to reconsider and leave the fancy name or do you want to have a static file name `with_audio.mp4`?

ammen99: I admit a fancier name is definitely useful, but it is beside the point when showing an example.

ammen99:
Thanks!

PR #86, opened 2020-04-17
Set codec parameters from context

soreau:
This fixes the following warning:

"Using AVStream.codec to pass codec parameters to muxers is deprecated,
use AVStream.codecpar instead"

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -192,6 +192,8 @@ void FrameWriter::init_video_stream()
     if (fmtCtx->oformat->flags & AVFMT_GLOBALHEADER)
         videoCodecCtx->flags |= AV_CODEC_FLAG_GLOBAL_HEADER;
 
+    avcodec_parameters_from_context(videoStream->codecpar, videoStream->codec);
+
     int ret;
     char err[256];
     if ((ret = avcodec_open2(videoCodecCtx, codec, &options)) < 0)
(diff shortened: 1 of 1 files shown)


soreau:
> I looked a bit on the internet, like [here](https://blog.mi.hdm-stuttgart.de/index.php/2018/03/21/livestreaming-with-libav-tutorial-part-2/) and [here](https://fossies.org/linux/ffmpeg/doc/examples/muxing.c).
> 
> Seems like we should use `avcodec_alloc_context3` instead of using the built-in `AVStream::codec` (and of course then deallocate the context at the end of the destructor, however it is done)

Doing this is a different idea than what this PR aims to do. You can do this if you want but it's well beyond the scope of fixing a warning.

ammen99:
> > 
> 
> Doing this is a different idea than what this PR aims to do. You can do this if you want but it's well beyond the scope of fixing a warning.

It is not a different idea. You'd simply have to set `videoCodecCtx` to the newly allocated one instead of using `AVStream::codec`. What I don't like with this is that it is not a fix, it is a workaround.

You can either fix it or close this PR.

soreau:
> > > 
> > 
> > 
> > Doing this is a different idea than what this PR aims to do. You can do this if you want but it's well beyond the scope of fixing a warning.
> 
> It is not a different idea.

Yes, it is.

> You'd simply have to set `videoCodecCtx` to the newly allocated one instead of using `AVStream::codec`.

As I said, you can patch it if you wish.

>What I don't like with this is that it is not a fix, it is a workaround.

Make of it what you will.

PR #88, opened 2020-04-19
Make pulseaudio support optional

soreau:


src/main.cpp
```diff
@@ -663,8 +683,10 @@ int main(int argc, char *argv[])
                 break;
 
             case 'a':
+#ifdef HAVE_PULSE
```
ammen99, line 686: I'd suggest printing an error in this case.

ammen99:
But needs a rebase ..

PR #93, opened 2020-05-03
main.cpp: add --list-video-codecs option

adtac:


adtac:
Later when support for audio codecs other than aac are added, we can change this to `--list-codecs` and print both video and audio codecs with the same command.

ammen99:
Can we not list these codecs with the `ffmpeg` command?

adtac:
fair enough, I didn't realise ffmpeg had such an option

feel free to merge/close

ammen99:
Okay, thank you for the PR but as already said, this lisy is already available by ffmpeg so no need to duplicate the functionality here.

PR #94, opened 2020-05-09
Quit wf-recorder when the recording target fails

progandy:
This can happen when the sdl muxer is used and the window is closed or when the target file reaches the filesystem limits.

    ./wf-recorder -c rawvideo -m sdl  -f pipe:mirror

I am not sure if this is the best solution, but it was the quickest fix.

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -601,7 +601,9 @@ void FrameWriter::finish_frame(AVCodecContext *enc_ctx, AVPacket& pkt)
         pending_mutex.unlock();
     }
 
-    av_interleaved_write_frame(fmtCtx, &pkt);
+    if (av_interleaved_write_frame(fmtCtx, &pkt) != 0) {
+        params.write_aborted_flag = true;
+    }
     av_packet_unref(&pkt);
 
     if (params.enable_audio)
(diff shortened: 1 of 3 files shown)


src/frame-writer.cpp
```diff
@@ -601,7 +602,9 @@ void FrameWriter::finish_frame(AVCodecContext *enc_ctx, AVPacket& pkt)
         pending_mutex.unlock();
     }
 
-    av_interleaved_write_frame(fmtCtx, &pkt);
+    if (!!av_interleaved_write_frame(fmtCtx, &pkt)) {
```
ammen99, line 605: what is the point of double negation? If you want something explicit, write `(... != 0)`

src/frame-writer.cpp
```diff
@@ -10,6 +10,7 @@
 #include <cstring>
 #include "averr.h"
 
+extern std::atomic<bool> exit_main_loop;
```
ammen99, line 13: Add a comment where this comes from (main.cpp in this case).

src/frame-writer.cpp
progandy: That is a habit I gained from reading kernel sources, I'll remove or change it.

src/frame-writer.cpp
progandy: Or maybe it is better to add a reference to that variable in FrameWriterParams instead?

PR #98, opened 2020-05-18
Add void linux install instructions

adrianisuru:
wf-recorder is also in the official repos of void linux


ammen99:
That's good to know, thanks.

Issue #100, opened 2020-06-10
How to change audio codec options? 


happysmash27:
I would like to record flac in 48000HZ. Is it currently possible to do this, and if not, would it be feasible for me, as a person with only basic C++ knowledge (but knowledge of C and the FFMPEG command line), to implement this before tomorrow?

happysmash27:
I'm surprised! I'm guessing you learned ffmpeg mostly as an API instead of a program? I use ffmpeg in the command line for many things (even some video editing occasionally), so the options I added that are not based on the current options are mostly based on ffmpeg ones, or are based on both. 

This is the command I used to use to record my video output: 
```
ffmpeg -f alsa -ac 2 -ar 44100 -i loopout -f alsa -i pulse -filter_complex amix=inputs=2 -f x11grab -s 1280x1024 -i :0.0 -r 60 -acodec pcm_s16le -vcodec libx264 -crf 0 -preset ultrafast output.mkv
ffmpeg -i output.mkv -vcodec libx264 final.avi
rm output.mkv
```
Looking back on it, I probably would have used different options today, but that is what I used back then. 

If it wasn't a conscious decision to use --codec instead of --vcodec, I would recommend the latter, perhaps preserving --codec as an alias to it for backwards compatibility, just to be a bit more consistent with ffmpeg and with having a codec option for both audio and video. Perhaps that should be a separate issue/pull request, though.

ammen99:
I'd rather not change what already exists, but maybe we can add vcodec for compatibility (in a separate PR). And yes, I learned mostly as an API, and I have almost zero experience with the actual ffmpeg command :)

happysmash27:
It took me a bit too long to implement the change of sample rate, but in testing, apparently it's actually _needed_ even with all the default formats. All my audio is out of sync, since it is 48000HZ but wf-recorder is interpreting it as 44100HZ. I already recorded the thing, and guess I will have to fix it in post. Hopefully I can fix the issue today after that.

happysmash27:
On second thought, after getting the sample format to work and more testing, I think that is an unrelated issue.

happysmash27:
I have made a pull request, but still have a function I am still trying to figure out in order to make it integrate best with the rest of the code. May I ask what the purpose of this function is: 

```
static enum AVSampleFormat get_codec_sample_fmt(AVCodec *codec)
{
    int i = 0;
    if (!codec->sample_fmts)
        return AV_SAMPLE_FMT_S16;
[...]

PR #102, opened 2020-06-19
Add support for changing audio codec options and video framerate

happysmash27:
This is a WIP fix for issue 100: https://github.com/ammen99/wf-recorder/issues/100. 

It also adds some general improvements like adding a video framerate option, and making options consistent between audio and video. 

The first stage just adds an audio codec option, but the second stage, after I add it, will add support for changing the audio codec, sample format, sample rate, and frame rate. It is not finished, however, as my change of sample format most likely breaks a function which I do not understand.

### DIFF
--- a/manpage/wf-recorder.1.scd
+++ b/manpage/wf-recorder.1.scd
@@ -20,7 +20,30 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
 *-c, --codec*
 	Specifies the codec of the video. Supports GIF output as well.
 
-	To modify codec parameters, use *-p <option_name>=<option_value>*
+*-r, --framerate*
+	Changes an approximation of the video framerate. The default value is *60*.
+
+*-x, --pixel-format*
+	Set the output pixel format. List available formats using *ffmpeg -pix_fmts*
+
+*-p, --codec-param*
+	Change the video codec parameters.
+	
+	*-p <option_name>=<option_value>*
+
+*-C, --acodec*
+	Specifies the codec of the audio.
+
+*-R, --sample-rate*
+	Changes the audio sample rate, in HZ. The default value is *44100*.
+
+*-X, --sample-format*
+	Set the output audio sample format. List available formats using *ffmpeg -sample_fmts*
+
+*-P, --acodec-param*
+	Change the audio codec parameters.
+	
+	*-P <option_name>=<option_value>*
 
 *-d, --device*
 	Selects the device to use when encoding the video.
@@ -39,9 +62,6 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
 
 *-m, --muxer*
 	Set the output format to a specific muxer instead of detecting it from the filename.
-
-*-x, --pixel-format*
-	Set the output pixel format. List available formats using *ffmpeg -pix_fmts*
 
 *-g, --geometry*
 	Selects a specific part of the screen.
--- a/manpage/wf-recorder.1.scd
+++ b/manpage/wf-recorder.1.scd
@@ -21,7 +21,7 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
 	Specifies the codec of the video. Supports GIF output as well.
 
 *-r, --framerate*
-	Changes an approximation of the video framerate. The default value is *60*.
+	Changes an approximation of the video framerate.
 
 *-x, --pixel-format*
 	Set the output pixel format. List available formats using *ffmpeg -pix_fmts*
@@ -35,7 +35,7 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
 	Specifies the codec of the audio.
 
 *-R, --sample-rate*
-	Changes the audio sample rate, in HZ. The default value is *44100*.
+	Changes the audio sample rate, in HZ.
 
 *-X, --sample-format*
 	Set the output audio sample format. List available formats using *ffmpeg -sample_fmts*
--- a/manpage/wf-recorder.1.scd
+++ b/manpage/wf-recorder.1.scd
@@ -24,7 +24,7 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
 	Changes an approximation of the video framerate.
 
 *-x, --pixel-format*
-	Set the output pixel format. List available formats using *ffmpeg -pix_fmts*
+	Set the output video pixel format. List available formats using *ffmpeg -pix_fmts*
 
 *-p, --codec-param*
 	Change the video codec parameters.
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -260,7 +260,7 @@ static enum AVSampleFormat convert_codec_sample_fmt(AVCodec *codec, std::string 
 	std::exit(-1);
     } else if (!codec->sample_fmts || check_fmt_available(codec, converted_fmt))
     {
-        std::cout << "Using sample format " << av_get_sample_fmt_name(converted_fmt) << std::endl;
+        std::cout << "Using sample format " << av_get_sample_fmt_name(converted_fmt) << " for audio codec " << codec->name << std::endl;
         return converted_fmt;
     } else
     {
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -10,8 +10,8 @@
 #include <cstring>
 #include "averr.h"
 
-#define FPS 60
-#define AUDIO_RATE 44100
+//#define FPS 60
+//#define AUDIO_RATE 44100
 
 class FFmpegInitialize
 {
@@ -94,6 +94,15 @@ void FrameWriter::load_codec_options(AVDictionary **dict)
     }
 }
 
+void FrameWriter::load_acodec_options(AVDictionary **dict)
+{
+    for (auto& opt : params.acodec_options)
+    {
+        std::cout << "Setting codec option: " << opt.first << "=" << opt.second << std::endl;
+        av_dict_set(dict, opt.first.c_str(), opt.second.c_str(), 0);
+    }
+}
+
 bool is_fmt_supported(AVPixelFormat fmt, const AVPixelFormat *supported)
 {
     for (int i = 0; supported[i] != AV_PIX_FMT_NONE; i++)
@@ -168,7 +177,8 @@ void FrameWriter::init_video_stream()
     videoCodecCtx = videoStream->codec;
     videoCodecCtx->width = params.width;
     videoCodecCtx->height = params.height;
-    videoCodecCtx->time_base = (AVRational){ 1, FPS };
+    std::cout << "Framerate: " << params.framerate << std::endl;
+    videoCodecCtx->time_base = (AVRational){ 1, params.framerate };
 
     if (params.bframes != -1)
         videoCodecCtx->max_b_frames = params.bframes;
@@ -218,11 +228,13 @@ static uint64_t get_codec_channel_layout(AVCodec *codec)
       return codec->channel_layouts[0];
 }
 
-static enum AVSampleFormat get_codec_sample_fmt(AVCodec *codec)
+static enum AVSampleFormat get_codec_sample_fmt(AVCodec *codec, AVSampleFormat requested_format)
 {
     int i = 0;
     if (!codec->sample_fmts)
-        return AV_SAMPLE_FMT_S16;
+        //return AV_SAMPLE_FMT_S16;
+        std::cout << "Using requested format: " << av_get_sample_fmt_name(requested_format) << std::endl;
+        return requested_format;
     while (1) {
         if (codec->sample_fmts[i] == -1)
             break;
@@ -235,6 +247,9 @@ static enum AVSampleFormat get_codec_sample_fmt(AVCodec *codec)
 
 void FrameWriter::init_audio_stream()
 {
+    AVDictionary *options = NULL;
+    load_codec_options(&options);
+    
     AVCodec* codec = avcodec_find_encoder_by_name(params.acodec.c_str());
     if (!codec)
     {
@@ -253,9 +268,9 @@ void FrameWriter::init_audio_stream()
     if (params.codec.c_str() == "aac"){
         audioCodecCtx->bit_rate = lrintf(128000.0f);
     }
-    audioCodecCtx->sample_fmt = get_codec_sample_fmt(codec);
+    audioCodecCtx->sample_fmt = get_codec_sample_fmt(codec, av_get_sample_fmt(params.sample_fmt.c_str()));
     audioCodecCtx->channel_layout = get_codec_channel_layout(codec);
-    audioCodecCtx->sample_rate = AUDIO_RATE;
+    audioCodecCtx->sample_rate = params.sample_rate;
     audioCodecCtx->time_base = (AVRational) { 1, 1000 };
     audioCodecCtx->channels = av_get_channel_layout_nb_channels(audioCodecCtx->channel_layout);
 
@@ -276,7 +291,7 @@ void FrameWriter::init_audio_stream()
         std::exit(-1);
     }
 
-    av_opt_set_int(swrCtx, "in_sample_rate", AUDIO_RATE, 0);
+    av_opt_set_int(swrCtx, "in_sample_rate", params.sample_rate, 0);
     av_opt_set_int(swrCtx, "out_sample_rate", audioCodecCtx->sample_rate, 0);
     av_opt_set_sample_fmt(swrCtx, "in_sample_fmt", AV_SAMPLE_FMT_FLT, 0);
     av_opt_set_sample_fmt(swrCtx, "out_sample_fmt", audioCodecCtx->sample_fmt, 0);
@@ -524,9 +539,10 @@ void FrameWriter::add_frame(const uint8_t* pixels, int64_t usec, bool y_invert)
 #define SRC_RATE 1e6
 #define DST_RATE 1e3
 
-static int64_t conv_audio_pts(SwrContext *ctx, int64_t in)
+static int64_t conv_audio_pts(SwrContext *ctx, int64_t in, int sample_rate)
 {
-    int64_t d = (int64_t) AUDIO_RATE * AUDIO_RATE;
+    //int64_t d = (int64_t) AUDIO_RATE * AUDIO_RATE;
+    int64_t d = (int64_t) sample_rate * sample_rate;
 
     /* Convert from audio_src_tb to 1/(src_samplerate * dst_samplerate) */
     in = av_rescale_rnd(in, d, SRC_RATE, AV_ROUND_NEAR_INF);
@@ -556,7 +572,8 @@ size_t FrameWriter::get_audio_buffer_size()
 void FrameWriter::add_audio(const void* buffer)
 {
     AVFrame *inputf = av_frame_alloc();
-    inputf->sample_rate    = AUDIO_RATE;
+    //inputf->sample_rate    = AUDIO_RATE;
+    inputf->sample_rate    = params.sample_rate;
     inputf->format         = AV_SAMPLE_FMT_FLT;
     inputf->channel_layout = AV_CH_LAYOUT_STEREO;
     inputf->nb_samples     = audioCodecCtx->frame_size;
@@ -571,7 +588,7 @@ void FrameWriter::add_audio(const void* buffer)
     outputf->nb_samples     = audioCodecCtx->frame_size;
     av_frame_get_buffer(outputf, 0);
 
-    outputf->pts = conv_audio_pts(swrCtx, INT64_MIN);
+    outputf->pts = conv_audio_pts(swrCtx, INT64_MIN, params.sample_rate);
     swr_convert_frame(swrCtx, outputf, inputf);
 
     send_audio_pkt(outputf);
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -235,10 +235,10 @@ static enum AVSampleFormat get_codec_sample_fmt(AVCodec *codec)
 
 void FrameWriter::init_audio_stream()
 {
-    AVCodec* codec = avcodec_find_encoder_by_name("aac");
+    AVCodec* codec = avcodec_find_encoder_by_name(params.acodec.c_str());
     if (!codec)
     {
-        std::cerr << "Failed to find the aac codec" << std::endl;
+        std::cerr << "Failed to find the given audio codec: " << params.acodec << std::endl;
         std::exit(-1);
     }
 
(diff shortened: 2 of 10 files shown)


happysmash27:
The version I have uploaded is perfectly functional for me (I have been using it to record for the past few days), but with other options it might not be. What is the purpose of the following function?: 

```
static enum AVSampleFormat get_codec_sample_fmt(AVCodec *codec)
{
    int i = 0;
    if (!codec->sample_fmts)
        return AV_SAMPLE_FMT_S16;
    while (1) {
        if (codec->sample_fmts[i] == -1)
            break;
        if (av_get_bytes_per_sample(codec->sample_fmts[i]) >= 2)
            return codec->sample_fmts[i];
        i++;
    }
    return codec->sample_fmts[0];
}
```
It seems to iterate through the list of sample formats a codec supports, from what I understand, but I do not understand it very well, especially as it uses unfamiliar syntax to me, without curly brackets around the first if statement, and with a library I do not know well in code. My current hastily-written re-implementation breaks the entire function in a way I hardly even understand just to get it to use the sample format I request, but this is not viable in production.

happysmash27:
I understand the function a bit better now, but still do not quite understand its purpose. Was it for future-proofing in case of an audio codec option? I have decided to work around it by having the sample format be NULL by default, and automatically determining the sample format only if no valid one is set.

ammen99:
@happysmash27 Sorry, I forgot about your question. The function was taken from another project's code because I myself am no expert on ffmpeg (the other project is wlstream). However, it just seems to try to pick a stereo format when possible. I think you should do the following:

1. If the user has explicitly specified audio format, then use that format only.
2. If the user hasn't explicitly specified the audio format, then use the function as it is in the master branch.

src/main.cpp
```diff
@@ -529,26 +547,20 @@ With no FILE, start recording the current screen.)");
...
 
   -h, --help                Prints this help screen.
 
   -l, --log                 Generates a log on the current terminal. Debug purposes.
 
-  -o, --output              Specify the output where the video is to be recorded.
-
-  -p, --codec-param         Change the codec parameters.
-                            -p <option_name>=<option_value>)");
+  -o, --output              Specify the output where the video is to be recorded.");
```
soreau, line 556: You're missing a closing parenthesis. Not you change `)");` to `");`

src/main.cpp
```diff
@@ -345,6 +345,7 @@ static void write_loop(FrameWriterParams params)
             if (params.enable_audio)
             {
                 pulseParams.audio_frame_size = frame_writer->get_audio_buffer_size();
+		pulseParams.sample_rate = params.sample_rate;
```
soreau, line 348: This has wrong indentation.

src/main.cpp
```diff
@@ -611,10 +627,15 @@ int main(int argc, char *argv[])
...
-        { "pixel-format",    required_argument, NULL, 'x' },
         { "geometry",        required_argument, NULL, 'g' },
         { "codec",           required_argument, NULL, 'c' },
         { "codec-param",     required_argument, NULL, 'p' },
+	{ "framerate",       required_argument, NULL, 'r' },
+        { "pixel-format",    required_argument, NULL, 'x' },
+	{ "acodec",          required_argument, NULL, 'C' },
+	{ "acodec-param",    required_argument, NULL, 'P' },
+	{ "sample-rate",     required_argument, NULL, 'R' },
+	{ "sample-format",   required_argument, NULL, 'X' },
```
soreau, line 638: These have mismatched indentation.

src/main.cpp
```diff
@@ -656,6 +673,26 @@ int main(int argc, char *argv[])
                 params.codec = optarg;
                 break;
 
+	    case 'r':
+	        params.framerate = atoi(optarg);
+	 	break;
+
+            case 'x':
```
soreau, line 680: Also indentation white space mismatch.

src/frame-writer.hpp
```diff
@@ -11,7 +11,7 @@
 #include <atomic>
 #include "config.h"
 
-#define AUDIO_RATE 44100
+//#define AUDIO_RATE 44100
```
soreau, line 14: You should remove items instead of commenting them, if they are not to be included in the final patch.

src/frame-writer.cpp
```diff
@@ -233,12 +243,37 @@ static enum AVSampleFormat get_codec_sample_fmt(AVCodec *codec)
     return codec->sample_fmts[0];
 }
 
+bool check_fmt_available(AVCodec *codec, AVSampleFormat fmt){
+    for (const enum AVSampleFormat *sample_ptr = codec -> sample_fmts; *sample_ptr != -1; sample_ptr++){
```
soreau, line 247:
```suggestion
    for (const enum AVSampleFormat *sample_ptr = codec -> sample_fmts; *sample_ptr != -1; sample_ptr++)
    {
```

src/frame-writer.cpp
```diff
@@ -233,12 +243,37 @@ static enum AVSampleFormat get_codec_sample_fmt(AVCodec *codec)
...
+	if (*sample_ptr == fmt){
```
soreau, line 248:
```suggestion
	if (*sample_ptr == fmt)
	{
```

happysmash27:
One more complicated approach I was taking, was that if the requested sample format is invalid, to use the automatic format instead. I found a perfect function to help do this in the following file, but didn't realise it was in example code, rather than in an include file, until I tried to use it and the function was not found: https://libav.org/documentation/doxygen/master/encode__audio_8c_source.html#l00040. 

In order to get around this, I have made another function that does the same thing but in c++, and with a different coding style. I may or may not just revert the entire section, though, despite working a while on it. I am wondering what the better outcome would be: The user inputs an invalid format, and the program continues on with a different format, or the user inputs an invalid format, and the program immediately dies when trying to record? 

If it stays on the error message route, it may make sense to make the error a bit more obvious with coloured text. Currently, it is very easy to miss, and it is easy to put the wrong option in for sample formats.

src/main.cpp
happysmash27: Thank you so much! That tiny error was driving me crazy and I had no idea what had changed to make the #endif statement no longer be valid.

soreau:
> I am wondering what the better outcome would be: The user inputs an invalid format, and the program continues on with a different format, or the user inputs an invalid format, and the program immediately dies when trying to record?
> 
> If it stays on the error message route, it may make sense to make the error a bit more obvious with coloured text. Currently, it is very easy to miss, and it is easy to put the wrong option in for sample formats.

I would expect the program to error if user inputs invalid format.

src/main.cpp
happysmash27: Yes, the coding style is a mess. My text editor kept using the wrong indentation, and only in the process of trying to change this to contribute to this project did I later realise it had a function to change the coding style used by the automatic indenter, and that this project had a completely different coding style than I am used to in other aspects as well. I am still in the process of cleaning all that up. 

...It seems my editor used tabs instead of spaces here, though, which looks exactly the same in my editor, but not in GitHub. So thank you for this tip too; I would be a lot less likely to catch it without it.

src/main.cpp
happysmash27: Hmm... I guess I need to figure out a way to tell Emacs to not use tabs at all. It is still doing it even when I try to set the coding style from existing code. Thank you for the notification; I couldn't see anything was off at all. Emacs has certainly given me a really big hassle contributing to this project with its strange automatic indents.

src/frame-writer.hpp
happysmash27: That is my plan. I am still cleaning up.

happysmash27:
I think you're right. Second-guessing the user is probably a bad idea.

happysmash27:
I'm having trouble figuring out how to input an integer for the default options in config.h.in (since before these now-configurable options where defined with #define statements in the header files themselves). I'm still searching for better information on how Meson works to do this.

happysmash27:
[This documentation](https://mesonbuild.com/Configuration.html) seems to suggest using conf_data.set() for this, including for strings, but the current meson_options.txt uses an option() syntax.

happysmash27:
[This documentation](https://mesonbuild.com/Build-options.html) for the option() function, on the other hand, seems to make no mention of config.h.in.

happysmash27:
I just figured out, that it appears I need to modify the meson.build file as well to make everything work. Now it makes much more sense!

happysmash27:
@ammen99 get_codec_sample_fmt being used to make audio dual-channel doesn't make sense, since from what I understand, sample format has nothing to do with the number of channels. It just specifies the bit depth of a sample and whether it is sampled as an integer or floating point number. Any sample format should be able to be used with any number of channels. However, after some testing, I believe the function is still necessary. 

If I change it to the following: 

```
static enum AVSampleFormat get_codec_auto_sample_fmt(AVCodec *codec)
{
    int i = 0;
    if (!codec->sample_fmts)
        return AV_SAMPLE_FMT_S16;
    while (1) {
	std::cout << "Sample format " << av_get_sample_fmt_name(codec->sample_fmts[i]) << " has " << av_get_bytes_per_sample(codec->sample_fmts[i]) << " bytes per sample" << std::endl;
        if (codec->sample_fmts[i] == -1)
            break;
        if (av_get_bytes_per_sample(codec->sample_fmts[i]) >= 2)
            return codec->sample_fmts[i];
        i++;
    }
    return codec->sample_fmts[0];
}
```

It will always only output one line, finding a format that satisfies the condition immediately both on the default audio codec and for flac: 

```
% ~/Gits/Fork/wf-recorder/build/wf-recorder -a -f recordtest_blank.mkv
Please select an output from the list to capture (enter output no.):
1. Name: DP-3 Description: Dell Inc. DELL 1907FPV CC3887B6AESZ (DP-3)
2. Name: HDMI-A-1 Description: Dell Inc. DELL 1907FPV CC3887B6AER2 (HDMI-A-1)
Enter output no.:1
selected region 0 0 0 0
Setting codec option: crf=20
Setting codec option: preset=ultrafast
Setting codec option: tune=zerolatency
Framerate: 60
Choosing pixel format yuv420p
[libx264 @ 0x7f5004001da0] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2
[libx264 @ 0x7f5004001da0] profile Constrained Baseline, level 4.2, 4:2:0, 8-bit
[libx264 @ 0x7f5004001da0] 264 - core 157 - H.264/MPEG-4 AVC codec - Copyleft 2003-2019 - http://www.videolan.org/x264.html - options: cabac=0 ref=1 deblock=0:0:0 analyse=0:0 me=dia subme=0 psy=1 psy_rd=1.00:0.00 mixed_ref=0 me_range=16 chroma_me=1 trellis=0 8x8dct=0 cqm=0 deadzone=21,11 fast_pskip=1 chroma_qp_offset=0 threads=16 lookahead_threads=16 sliced_threads=1 slices=16 nr=0 decimate=1 interlaced=0 bluray_compat=0 constrained_intra=0 bframes=0 weightp=0 keyint=250 keyint_min=25 scenecut=0 intra_refresh=0 rc=crf mbtree=0 crf=20.0 qcomp=0.60 qpmin=0 qpmax=69 qpstep=4 ip_ratio=1.40 aq=0
Setting codec option: crf=20
Setting codec option: preset=ultrafast
Setting codec option: tune=zerolatency
Sample format fltp has 4 bytes per sample
Output #0, matroska, to 'recordtest_blank.mkv':
    Stream #0:0: Unknown: none (libx264)
    Stream #0:1: Unknown: none (aac)
[matroska @ 0x7f5004000d90] Using AVStream.codec to pass codec parameters to muxers is deprecated, use AVStream.codecpar instead.
[matroska @ 0x7f5004000d90] Using AVStream.codec to pass codec parameters to muxers is deprecated, use AVStream.codecpar instead.
Using PulseAudio device: default
^C[libx264 @ 0x7f5004001da0] frame I:1     Avg QP:17.00  size:222742
[libx264 @ 0x7f5004001da0] frame P:2     Avg QP:23.50  size:   262
[libx264 @ 0x7f5004001da0] mb I  I16..4: 100.0%  0.0%  0.0%
[libx264 @ 0x7f5004001da0] mb P  I16..4:  0.0%  0.0%  0.0%  P16..4:  0.0%  0.0%  0.0%  0.0%  0.0%    skip:100.0%
[libx264 @ 0x7f5004001da0] coded y,uvDC,uvAC intra: 39.0% 33.5% 28.9% inter: 0.0% 0.0% 0.0%
[libx264 @ 0x7f5004001da0] i16 v,h,dc,p: 49% 34% 16%  1%
[libx264 @ 0x7f5004001da0] i8c dc,h,v,p: 72% 12% 15%  1%
[libx264 @ 0x7f5004001da0] kb/s:35722.40
[aac @ 0x7f500440be30] Qavg: 65536.000
% ~/Gits/Fork/wf-recorder/build/wf-recorder -a -C flac -f recordtest_blank.mkv
Please select an output from the list to capture (enter output no.):
1. Name: DP-3 Description: Dell Inc. DELL 1907FPV CC3887B6AESZ (DP-3)
2. Name: HDMI-A-1 Description: Dell Inc. DELL 1907FPV CC3887B6AER2 (HDMI-A-1)
Enter output no.:1
selected region 0 0 0 0
Setting codec option: crf=20
Setting codec option: preset=ultrafast
Setting codec option: tune=zerolatency
Framerate: 60
Choosing pixel format yuv420p
[libx264 @ 0x7f98d4001da0] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2
[libx264 @ 0x7f98d4001da0] profile Constrained Baseline, level 4.2, 4:2:0, 8-bit
[libx264 @ 0x7f98d4001da0] 264 - core 157 - H.264/MPEG-4 AVC codec - Copyleft 2003-2019 - http://www.videolan.org/x264.html - options: cabac=0 ref=1 deblock=0:0:0 analyse=0:0 me=dia subme=0 psy=1 psy_rd=1.00:0.00 mixed_ref=0 me_range=16 chroma_me=1 trellis=0 8x8dct=0 cqm=0 deadzone=21,11 fast_pskip=1 chroma_qp_offset=0 threads=16 lookahead_threads=16 sliced_threads=1 slices=16 nr=0 decimate=1 interlaced=0 bluray_compat=0 constrained_intra=0 bframes=0 weightp=0 keyint=250 keyint_min=25 scenecut=0 intra_refresh=0 rc=crf mbtree=0 crf=20.0 qcomp=0.60 qpmin=0 qpmax=69 qpstep=4 ip_ratio=1.40 aq=0
Setting codec option: crf=20
Setting codec option: preset=ultrafast
Setting codec option: tune=zerolatency
Sample format s16 has 2 bytes per sample
Output #0, matroska, to 'recordtest_blank.mkv':
    Stream #0:0: Unknown: none (libx264)
    Stream #0:1: Unknown: none (flac)
[matroska @ 0x7f98d4000d90] Using AVStream.codec to pass codec parameters to muxers is deprecated, use AVStream.codecpar instead.
[matroska @ 0x7f98d4000d90] Using AVStream.codec to pass codec parameters to muxers is deprecated, use AVStream.codecpar instead.
Using PulseAudio device: default
^C[libx264 @ 0x7f98d4001da0] frame I:1     Avg QP:17.00  size:222743
[libx264 @ 0x7f98d4001da0] frame P:3     Avg QP:22.33  size:   266
[libx264 @ 0x7f98d4001da0] mb I  I16..4: 100.0%  0.0%  0.0%
[libx264 @ 0x7f98d4001da0] mb P  I16..4:  0.0%  0.0%  0.0%  P16..4:  0.1%  0.0%  0.0%  0.0%  0.0%    skip:99.9%
[libx264 @ 0x7f98d4001da0] coded y,uvDC,uvAC intra: 39.0% 33.5% 28.9% inter: 0.0% 0.0% 0.0%
[libx264 @ 0x7f98d4001da0] i16 v,h,dc,p: 49% 34% 16%  1%
[libx264 @ 0x7f98d4001da0] i8c dc,h,v,p: 72% 12% 15%  1%
[libx264 @ 0x7f98d4001da0] kb/s:26824.92

```
The interesting thing, though, is that it chooses fltp for the default codec, and the default codec does not actually appear to support what has been set as the default format if it is forced: 

```
% ~/Gits/Fork/wf-recorder/build/wf-recorder -a -X s16 -f recordtest_blank.mkv
Please select an output from the list to capture (enter output no.):
1. Name: DP-3 Description: Dell Inc. DELL 1907FPV CC3887B6AESZ (DP-3)
2. Name: HDMI-A-1 Description: Dell Inc. DELL 1907FPV CC3887B6AER2 (HDMI-A-1)
Enter output no.:1
selected region 0 0 0 0
Setting codec option: crf=20
Setting codec option: preset=ultrafast
Setting codec option: tune=zerolatency
Framerate: 60
Choosing pixel format yuv420p
[libx264 @ 0x7f7a0c001da0] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2
[libx264 @ 0x7f7a0c001da0] profile Constrained Baseline, level 4.2, 4:2:0, 8-bit
[libx264 @ 0x7f7a0c001da0] 264 - core 157 - H.264/MPEG-4 AVC codec - Copyleft 2003-2019 - http://www.videolan.org/x264.html - options: cabac=0 ref=1 deblock=0:0:0 analyse=0:0 me=dia subme=0 psy=1 psy_rd=1.00:0.00 mixed_ref=0 me_range=16 chroma_me=1 trellis=0 8x8dct=0 cqm=0 deadzone=21,11 fast_pskip=1 chroma_qp_offset=0 threads=16 lookahead_threads=16 sliced_threads=1 slices=16 nr=0 decimate=1 interlaced=0 bluray_compat=0 constrained_intra=0 bframes=0 weightp=0 keyint=250 keyint_min=25 scenecut=0 intra_refresh=0 rc=crf mbtree=0 crf=20.0 qcomp=0.60 qpmin=0 qpmax=69 qpstep=4 ip_ratio=1.40 aq=0
Setting codec option: crf=20
Setting codec option: preset=ultrafast
Setting codec option: tune=zerolatency
Codec aac does not support sample format s16
```

So, it appears that its purpose is to find a supported sample format for a codec with a bit depth of at least 16, because AV_SAMPLE_FMT_U8 (which is 8-bit) exists, but almost no one in their right mind would want to record 8-bit audio instead of 16-bit or more.

happysmash27:
The message for the audio codec chosen is different depending on whether the user specifies it or wf-recorder auto-chooses it on purpose, by the way. If the computer chooses it, it matches with the "choosing pixel format" message, but if the user specifies it, it will say "using" instead, since the program didn't choose anything.

happysmash27:
The message for the audio codec chosen is different depending on whether the user specifies it or wf-recorder auto-chooses it on purpose, by the way. If the computer chooses it, it matches with the "choosing pixel format" message, but if the user specifies it, it will say "using" instead, since the program didn't choose anything.

happysmash27:
Ideally, I think wf-recorder would be able to determine framerate, sample rate, and sample format automatically based on refresh rate of the monitor, sample rate of the audio input, and sample format of the audio input, but that is a project for another pull request.

happysmash27:
I think everything is finally ready for review! So far, I can't find anything obvious left to fix.

manpage/wf-recorder.1.scd
```diff
@@ -20,7 +20,30 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
 *-c, --codec*
 	Specifies the codec of the video. Supports GIF output as well.
 
-	To modify codec parameters, use *-p <option_name>=<option_value>*
+*-r, --framerate*
```
ammen99, line 23: I think you should stress that this is just an approximation and doesn't really change the refresh rate, or does it?

src/main.cpp
```diff
@@ -705,6 +741,20 @@ int main(int argc, char *argv[])
                 }
                 break;
 
+	    case 'P':
+                param = optarg;
```
ammen99, line 745: This code is shared with the video parameter parsing, can we have a separate function like:

```
static void parse_codec_option(std::map<std::string, std::string>& options, std::string param)
{
...
}
```

src/main.cpp
```diff
@@ -656,6 +672,26 @@ int main(int argc, char *argv[])
...
+            case 'x':
+                params.pix_fmt = optarg;
+                break;
+
+	    case 'C':
+	        params.acodec = optarg;
+	        break;
+
+	    case 'R':
+	        params.sample_rate = atoi(optarg);
```
ammen99, line 688: We should at least print what values we have parsed for sample_rates, because otherwise if the user makes a typo they will have a hard time spotting it.

src/frame-writer.cpp
```diff
@@ -250,10 +289,18 @@ void FrameWriter::init_audio_stream()
     }
 
     audioCodecCtx = audioStream->codec;
-    audioCodecCtx->bit_rate = lrintf(128000.0f);
-    audioCodecCtx->sample_fmt = get_codec_sample_fmt(codec);
+    if (params.codec == "aac"){
```
ammen99, line 292: I looked into it again, and it seems that we do not need to set the bit rate at all. You can delete this if altogether.

src/frame-writer.cpp
```diff
@@ -250,10 +289,18 @@ void FrameWriter::init_audio_stream()
...
+        audioCodecCtx->bit_rate = lrintf(128000.0f);
+    }
+    if (params.sample_fmt.size() == 0){
```
ammen99, line 295: style: `{` on a new line

ammen99 (changes requested):
I left a few comments, but the overall implementation is quite good.

PS: I'm really sorry for the long wait, this PR slipped off my mind.

binex-dsk:
if you could finish doing this that would be very cool

keithbowes:
How's this coming along?  It would be great to be able not to have to either record with illegal patent-encumbered formats or (like I do) use a script to simultaneously record the video in VP9 with wf-recorder and the audio in opus with ffmpeg and then have ffmpeg merge the files.

manpage/wf-recorder.1.scd
happysmash27: It's been a while since I made this, but IIRC this was because the actual frame rate sometimes isn't actually achieved? Perhaps it would be better described as target frame rate. It attempts to reach it, but if the encoding is not fast enough, in practice it will be a lower frame rate. 

It should also be noted that when recording at a high frame rate, and also having heavy encoding, there can be desyncing issues if it can't actually write frames fast enough. I am not currently sure what is needed to fix this, other than by using easier encoding and recording at a lower framerate. If Wayland is anything like X11, it is also possible that actually retrieving the frames could be a bottleneck.

ammen99:
> How's this coming along? It would be great to be able not to have to either record with illegal patent-encumbered formats or (like I do) use a script to simultaneously record the video in VP9 with wf-recorder and the audio in opus with ffmpeg and then have ffmpeg merge the files.

The PR has to be rebased on top of master. I think the original author might not respond soon, so feel free (you or anybody else) to pick this up.

happysmash27:
Indeed, I have two or three super high-priority tasks while this is only medium-low priority, although still bothers me a LOT since people have been waiting so long hence my reply 29 days ago when I started trying to re-base... then realised that to remember properly what everything does I might actually have to test which would require a bit of a shift since I am not currently loaded into Sway and would have to close everything I currently have open (in LXDE) to avoid the potential for that catastrophic graphics bug where AMDGPU freezes up when too much is going on. It will most likely be July when I may be able to do things more properly, as I want to finish the project I am currently focusing on most by June 17th.

binex-dsk:
I  believe this is what is known as a "bruh moment"

ammen99:
This PR was updated and merged in #177, closing, thanks nonetheless :)

PR #107, opened 2020-07-18
wf-recorder in the official Arch repos now

basedtho:
https://www.archlinux.org/packages/community/x86_64/wf-recorder/


ammen99:
Nice to know! Thanks.

PR #117, opened 2020-10-07
docs: adds useful bits from README to man page.

markstos:
Knowing how to STOP `wf-recorder` is critical and was missing from the
man page.

The README had several other useful tips as well which werent in the man
page.

You shouldn't have to read both to get complete docs.

### DIFF
--- a/manpage/wf-recorder.1.scd
+++ b/manpage/wf-recorder.1.scd
@@ -8,6 +8,10 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
 
 *wf-recorder* [options...] -f [file]
 
+In its simplest form, run `wf-recorder` to start recording and use Ctrl+C to stop.
+This will create a file called `recording.mp4` in the current working directory using
+the default codec.
+
 # OPTIONS
 *-a, --audio [DEVICE]*
 	Starts recording the screen with audio.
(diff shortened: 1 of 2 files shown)


manpage/wf-recorder.1.scd
```diff
@@ -74,6 +80,44 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
 *wf-recorder* is a tool built to record your screen on Wayland compositors.
 It makes use of wlr-screencopy for capturing video and FFmpeg for encoding it.
 
+To select a specific part of the screen you can either use `-g <geometry>`, or
```
ammen99, line 83:
```suggestion
# EXAMPLES

To select a specific part of the screen you can either use `-g <geometry>`, or
```

manpage/wf-recorder.1.scd
```diff
@@ -74,6 +80,44 @@ wf-recorder - A simple screen recording program for wlroots-based compositors
...
+use [slurp](https://github.com/emersion/slurp) for interactive selection of the
```
ammen99, line 84:
```suggestion
use https://github.com/emersion/slurp for interactive selection of the
```

This is because man does not how links like github does.

markstos: Thanks. I meant to unlink this. I read `man scdoc` and noticed it doesn't seem to support Markdown hyperlinks.

ammen99 (approved):
LGTM, thanks!

PR #120, opened 2020-10-30
Fix wrong command to discover audio input devices, advertised in --help

soreau:


soreau:
#121 is better than this one because it doesn't show the <> around the device name, which isn't supposed to be included with the command. I will update the wiki with this information.

PR #121, opened 2020-10-30
Replace `sinks` with `sources` in help for --audio

### LINKED ISSUE #119: Help for --audio might be incorrect. Shouldn't it be `pactl list sources`?

cauebs:
I understand this is what's intended. I don't have any deep understanding of pulseaudio, but It's what worked for me and, as pointed out to me in #119, what is actually in this repo's wiki.

Closes #119


ammen99 (approved):
Nice catch!

PR #126, opened 2020-11-28
use pregenerated mdoc man page

epsilon-0:
gets rid of scdoc dependency

Signed-off-by: Aisha Tammy <floss@bsd.ac>

---

everytime scdoc is used a banana goes bad :broken_heart:


epsilon-0:
To view the manpage locally, just download it and do `man -l wf-recorder.1`

ammen99:
NACK. I prefer being able to write the manpage in a human-readable way without needing to know that weird syntax.

What is the problem with scdoc?

epsilon-0:
There is a lot of metadata lost using scdoc that is used in standard man pages.
Cross referencing to another man page, inside the text (not in See also). Cross referencing to another section inside the article.
Annotating the words as referencing the argument for the option, so that people know we are talking about the option and not a general thing, like
https://github.com/ammen99/wf-recorder/pull/126/files#diff-c251b89ae13a369d261b400fda6874b34b37ca0655af4aeedde2765501d086faR86
All of this metadata is used in the final displaying of the man page.
For example, here is the man page when generated with scdoc - 
https://i.imgur.com/Z1g0HnO.webp
And here it is when using the standard mdoc language - 
https://i.imgur.com/gzT51TV.webp

There's already a lot more meta information that is visible in the man page, even in the terminal.
Not to mention that you can render the man page as an html with proper links for headers, options, etc
https://i.imgur.com/ZsepJwb.webp
I've attached the default html generated man page, if you want to take a look (extension is `.txt` to get around github not allowing upload, rename to `.html` to view in browser after download)
[wf-recorder.1.txt](https://github.com/ammen99/wf-recorder/files/5610573/wf-recorder.1.txt)

ammen99:
@epsilon-0 Alright, your two screenshots won me over to mdoc :) Thanks for the PR and for the detailed explanation.

PR #141, opened 2021-02-20
don't ask for overwrite for the char device

### LINKED ISSUE #140: non-interactive capture to the /dev/video
with #115 implemented I can't bind ```wf-recorder -g "$(slurp)" --muxer=v4l2 --codec=rawvideo --file=/dev/video0 --pixel-format yuyv422``` to the hot key as /dev/video0 alway exist, and it requires confirmation.

Could we have a force option or maybe check in `stat` if that a special file/device?

denisok:
`/dev/videoX` should always exists for the `--muxer=v4l2`, don't ask permission to override it. Fixes #140 .

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -577,7 +577,7 @@ void handle_graceful_termination(int)
 static bool user_specified_overwrite(std::string filename)
 {
     struct stat buffer;   
-    if (stat (filename.c_str(), &buffer) == 0)
+    if (stat (filename.c_str(), &buffer) == 0 && !S_ISCHR(buffer.st_mode))
     {
         std::string input;
         std::cout << "Output file \"" << filename << "\" exists. Overwrite? Y/n: ";


denisok:
@soreau please review

soreau:
> `/dev/videoX` should always exists for the `--muxer=v4l2`, don't ask permission to override it. Fixes #140 .

If this is the case, wouldn't it make better sense to check if the [muxer](https://github.com/ammen99/wf-recorder/blob/master/src/frame-writer.hpp#L55) is "v4l2"?

denisok:
@soreau for that we probably need enum and validation of params and etc.
Logic that is implemented is more straightforward - ask for regular files, no sense to ask for the devices.

I don't know `libav` at all, but I would think there are more situation possible when `-f` isn't regular file.

soreau:
> @soreau for that we probably need enum and validation of params and etc.

My suggestion is to pass `params.mutex` to `user_specified_overwrite()` in addition to `params.file` here. Then in the function, you can check `mutex == "v4l2"` instead of checking if the file is a character device file. No need for an enum or anything.

> Logic that is implemented is more straightforward - ask for regular files, no sense to ask for the devices.
> 
> I don't know `libav` at all, but I would think there are more situation possible when `-f` isn't regular file.

I don't know of other cases where you would be passing an irregular existing file type to wf-recorder.

src/main.cpp
```diff
@@ -403,10 +403,10 @@ void handle_graceful_termination(int)
     exit_main_loop = true;
 }
 
-static bool user_specified_overwrite(std::string filename)
+static bool user_specified_overwrite(std::string filename, std::string muxername)
 {
     struct stat buffer;   
-    if (stat (filename.c_str(), &buffer) == 0)
+    if (stat (filename.c_str(), &buffer) == 0 && muxername != "v4l2")
```
ammen99, line 409: I'm thinking, can we not just check whether the file is a character device? Why do we need to special case v4l2?

ammen99:
@denisok Looking at this again, I think I agree with your original proposal (e.g. check whether it is a character device), not only whether the muxer is v4l2. I am also not aware of any other use-case, but there is nothing inherently special about v4l2.

ammen99 (approved):
Thanks and sorry for the long delay!

PR #151, opened 2021-05-18
Update README.md

dagbdagb:
Some additions to README w.r.t. Microsoft Teams. And gentoo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,12 @@ Fedora users can install from rpmfusion-free-updates. First [enable rpmfusion](h
 sudo dnf install wf-recorder
 ```
 
+## Gentoo Linux
+You may or may not have to adjust the USE-flags for media-video/ffmpeg prior to emerging wf-recorder. (pulseaudio, v4l, libv4l). And if you have apulse installed, it has to go. Note: The current wf-recorder code has pulseaudio as an optional feature, but the gentoo ebuild does not reflect this at the time of writing.
+```
+emerge wf-recorder
+```
+
 ## From Source
 ### Install Dependencies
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,7 +57,7 @@ git clone https://github.com/ammen99/wf-recorder.git && cd wf-recorder
 meson build --prefix=/usr --buildtype=release
 ninja -C build
 ```
-Optionally configure with `-Ddefault_codec='codec'`. The default is libx264. Now you can just run `./build/wf-recorder` or install it with `sudo ninja -C build install`.
+Optionally configure with `-Ddefault_codec='codec'`, or modify meson_options.txt. The default codec is libx264. Now you can just run `./build/wf-recorder` or install it with `sudo ninja -C build install`.
```
ammen99, line 60: I don't see why modifying `meson_options.txt` is a good option at all.

README.md
```diff
@@ -79,6 +85,20 @@ To set a specific output format, use the `--muxer` option. For example, to outpu
 wf-recorder --muxer=v4l2 --codec=rawvideo --file=/dev/video2
 ```
 
+Microsoft Teams appears to be a bit picky. Install v4l2loopback.ko like this:
```
ammen99, line 88: Seems like a good write-up, but I do not think it really fits into the README, maybe we can add this to the wiki and add a link from the readme to it?

ammen99 (changes requested):

README.md
dagbdagb: Frankly, I had no clue whatsoever what I should put in front of  `-Ddefault_codec=`, so I just fired up vim.....

README.md
dagbdagb: Absolutely. Can give it a try later, ran out of steam just now.  :-)

ammen99 (changes requested):
A little late, but looking over this, it will be better to put this information in the wiki, in a new page with workarounds for individual applications like MS Teams. Closing the PR, but feel free to extend the wiki.

README.md
eli-schwartz: BTW you can run `meson configure` and it will print out a description of all build options and what their possible values are. :)

Issue #155, opened 2021-09-01
Support for high bit depth screencopy buffers


mstoeckl:
wlroots will eventually be able to render to 10-bit or higher internal buffers. (See https://github.com/swaywm/wlroots/issues/1378.) † The internal buffers determine what buffer formats are requested by `zwlr_screencopy_frame_v1.buffer`;  so for example, with 10-bit buffers, `XBGR2101010` or `ABGR2101010` would be sent; and with half-float formats, `ABGR16161616F` or `XBGR16161616F` would be sent.

It will probably be a long time before support for these formats will be needed; but I am making this issue now, as this task may be difficult. Currently, ffmpeg does not have any direct equivalents for the above formats, so it may be necessary to add a format conversion step to wf-recorder. (For example, the closest thing to WL_SHM_FORMAT_XBGR2101010 is AV_PIX_FMT_X2RGB10LE, which has red and blue channels switched.)

† ~In case you want to test wlroots with higher bit depth earlier, I have made experimental branches for [10](https://github.com/mstoeckl/wlroots/tree/plane-10) and [16float](https://github.com/mstoeckl/wlroots/tree/plane-16f) internal depth; 16 bit depth may become possible in a few months.~ Edit: wlroots now supports changing the render buffer format; I made a Sway PR, https://github.com/swaywm/sway/pull/6475, which adds a setting to control this.

ammen99:
Wouldn't it make more sense to add support for these formats to ffmpeg, at least for conversion?

mstoeckl:
> Wouldn't it make more sense to add support for these formats to ffmpeg, at least for conversion?

Yes, that would be ideal.

mstoeckl:
I've filed a ticket about supporting XBGR2101010 at https://trac.ffmpeg.org/ticket/9421 , and have attached a (first draft) patch to add the format. (I'll try to submit a proper version to ffmpeg-devel later.) See https://github.com/mstoeckl/wf-recorder/commit/dca878c0022ed2d3a81f8daafde43f514d0e6406 for the changes needed to add XBGR2101010 support, plus changes needed to build against ffmpeg git master.

(It doesn't look like ffmpeg has any code at all to convert half-float formats, so I'll leave figuring out XBGR16161616F and ABGR16161616F support to future people.)

mstoeckl:
[...]

PR #157, opened 2021-11-20
Fix build with latest (git) FFmpeg

mstoeckl:
This makes it possible to compile `wf-recorder` using FFmpeg built from git. It should still be possible to build `wf-recorder` with earlier versions of FFmpeg, although I only checked with FFmpeg 4.4.

This change is necessary if wf-recorder is eventually to use the AV_PIX_FMT_X2BGR10LE format, which would be needed for #155.

FFmpeg 4.5 will probably come out in a few months, if the recent trend of 9-12 months between releases continues.

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -15,6 +15,8 @@
 static const AVRational US_RATIONAL{1,1000000} ;
 #define AUDIO_RATE 44100
 
+// av_register_all was deprecated in 58.9.100, removed in 59.0.100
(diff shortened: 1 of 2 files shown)


ammen99:
Needs rebasing on top of the filters branch

mstoeckl:
> Needs rebasing on top of the filters branch (which was just merged into master)
Done.

ammen99 (approved):
Thanks!

PR #161, opened 2022-01-13
Add support for 565 and 2101010 formats

mstoeckl:
This PR makes it possible to record the screen when the compositor provides 5-bit-per-channel or 10-bpc screenshot data instead of the usual 8-bpc.

This PR partially addresses #155 ; it does not add 16-unsigned-bpc and 16-float-bpc formats, since I don't have 16-unsigned working properly in mesa and wlroots yet, and FFmpeg does not yet support 16-float .

While the changes here are relatively simple, actually testing that the new formats work is tricky.

The 2101010 formats are only available if you use a recent (~~git~~ now 5.0) version of FFmpeg. To test the X2BGR format, install Sway ~~from git (or wait until 1.7 is released)~~ version 1.7, and run the `swaymsg output '*' render_bit_depth 10` command to increase the bit depth at which the screen is rendered.  Then `wf-recorder` should be given screencopy buffers with 10-bpc data.

To test the RGB565 format, build Sway from the [`mstoeckl/sway/extra-bit-depths`](https://github.com/mstoeckl/sway/tree/extra-bit-depths) branch, run `swaymsg output '*' render_bit_depth 5`; and then run `wf-recorder`.

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -92,6 +92,14 @@ AVPixelFormat FrameWriter::get_input_format()
         return AV_PIX_FMT_RGB565LE;
     case INPUT_FORMAT_BGR565:
         return AV_PIX_FMT_BGR565LE;
+#if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(56, 55, 100)
+    case INPUT_FORMAT_X2RGB10:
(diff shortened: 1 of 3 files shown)


ammen99 (approved):
I failed to compile your branch of Sway, but the code does not look too complicated so LGTM. Thanks!

PR #168, opened 2022-03-17
Fix: Crash when output is switched on

pmkap:
Currently, `chosen_output` is a pointer to an element in the `available_outputs` vector. When an output is plugged in or switched on, this vector grows, is reallocated and `chosen_output` becomes invalid, resulting in a segmentation fault.

Fixing this by using `std::list` instead of `std::vector` for
`available_outputs`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2,6 +2,7 @@
 #define _POSIX_C_SOURCE 199309L
 #include <iostream>
 
+#include <list>
 #include <string>
 #include <thread>
 #include <mutex>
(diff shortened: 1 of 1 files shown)


ammen99:
I think a simpler solution would be to just stop adding new outputs once we have started recording.

pmkap:
> I think a simpler solution would be to just stop adding new outputs once we have started recording.

Yeah I suppose so.

Even simpler would be to just reserve 10 or so outputs.

But IMO, a linked list is the better container for this case.

src/main.cpp
```diff
@@ -463,10 +464,16 @@ static wf_recorder_output* choose_interactive()
     fflush(stdout);
 
     int choice;
-    if (scanf("%d", &choice) != 1 || choice > (int)available_outputs.size() || choice <= 0)
-        return nullptr;
+    scanf("%d", &choice);
 
-    return &available_outputs[choice - 1];
+    i = 1;
```
ammen99, line 469: I'd rather use `std::advance` here, gives you better readability. Also, you'd need to keep the check above.

pmkap: Ah yeah this is a lot nicer :)

ammen99 (approved):
Thanks!

PR #171, opened 2022-04-20
Note that wf-recorder is available in Debian

gwolf:
wf-recorder has been available in Debian for three years already, and is part of the current stable release.
Thanks!


ammen99 (approved):
That's great, thanks for the PR ;)

PR #176, opened 2022-07-07
Set sample rate to 48000 to match Pipewire defaults

### LINKED ISSUE #59: audio sync is completely off
Recording video works great, but audio is ahead of it's time. Any way to fix this? I'm using Pulseaudio 13.0.

llyyr:
Alternative solution is to add 44100 in `default.clock.allowed-rates` in pipewire.conf. Better solution is to wait for #102 to be merged

Fixes #59 and other similar issues

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -13,7 +13,7 @@
 
 
 static const AVRational US_RATIONAL{1,1000000} ;
-#define AUDIO_RATE 44100
+#define AUDIO_RATE 48000
 
 // av_register_all was deprecated in 58.9.100, removed in 59.0.100
 #if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(59, 0, 100)
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -13,7 +13,6 @@
 
 
 static const AVRational US_RATIONAL{1,1000000} ;
-#define AUDIO_RATE 48000
 
 // av_register_all was deprecated in 58.9.100, removed in 59.0.100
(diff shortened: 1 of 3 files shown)


ammen99 (commented):
LGTM, it would be good if more people can test this to make sure we do not break anything. I'll try it later today myself.

src/frame-writer.cpp
```diff
@@ -13,7 +13,7 @@
 
 
 static const AVRational US_RATIONAL{1,1000000} ;
-#define AUDIO_RATE 44100
+#define AUDIO_RATE 48000
```
soreau, line 16: While we're at it, we should remove this #define since it's already defined in the header this file includes.

src/pulse.cpp
```diff
@@ -20,7 +20,7 @@ PulseReader::PulseReader(PulseReaderParams _p)
     pa_sample_spec sample_spec =
     {
         .format = PA_SAMPLE_FLOAT32LE,
-        .rate = 44100,
+        .rate = 48000,
```
soreau, line 23: ..and use the AUDIO_RATE define here instead of hard coding it.

soreau (commented):
This should have been a one-line change to the header but this is a good opportunity to clean up.

ammen99:
Is this necessary at all if #177 is merged?

llyyr:
Nope, this should be closed in favor of #177. I could change the default sample rate in that branch to 48k as well though

ammen99:
Closing as #177 was merged as a much better solution, thanks nonetheless ;)

PR #177, opened 2022-07-07
Support for configuring audio codec and video framerate

llyyr:
Rebase #102


manpage/wf-recorder.1.scd
```diff
@@ -0,0 +1,99 @@
+wf-recorder(1)
```
ammen99, line 1: We already have a manpage, the audio options should be added to it instead of creating a new manpage.

src/frame-writer.cpp
```diff
@@ -655,7 +704,8 @@ size_t FrameWriter::get_audio_buffer_size()
 void FrameWriter::add_audio(const void* buffer)
 {
     AVFrame *inputf = av_frame_alloc();
-    inputf->sample_rate    = AUDIO_RATE;
+    //inputf->sample_rate    = AUDIO_RATE;
```
ammen99, line 707: The comment should be removed as this code is not necessary any more.

manpage/wf-recorder.1.scd
llyyr: Ah, it's from the prev PR I didn't realize there was a new one

src/main.cpp
```diff
@@ -785,6 +820,20 @@ int main(int argc, char *argv[])
                 use_damage = false;
                 break;
 
+            case 'P':
+                param = optarg;
```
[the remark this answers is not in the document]
llyyr: Ah I'd appreciate if you could handle this one, I don't write c++ often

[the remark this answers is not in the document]
llyyr: Never mind, it was simple enough

ammen99 (approved):
LGTM now, thanks @llyyr for finishing up the work and thanks to @happysmash27 for the original work!

llyyr:
Are there any plans to make a new release? I think it'd be a good idea

ammen99:
> Are there any plans to make a new release? I think it'd be a good idea

Yeah, but let's first give people some time to test this.

apiraino:
Thanks for everyone working on this! Please let me know if I should file a separate issue.

Which are the possible values for `-c, --codec` and the new `-C, --audio-codec`? I've tried using `--audio-codec=mp3` or `--audio-codec=ogg` but they both don't work. Maybe a couple of examples in the wiki to get people started?

If this depends on the local installation of ffmpeg, how can I discover what is supported on my machine (`ffmpeg -codecs` show *all* supported codecs from ffmpeg)?

PR #178, opened 2022-07-09
Remove unused variable

WhyNotHugo:
Mostly squelching a built-time warning.


llyyr:
`    std::string param;` is unused and should be removed as well.

ammen99 (approved):
The code compiles with this patch, so lgtm :)
Thanks!

PR #180, opened 2022-07-26
docs: fix help and examples regarding -a & --audio

llyyr:


apiraino:
@llyyr thanks for encoding these useful info in the README.

Anything useful to salvage from [my comment](https://github.com/ammen99/wf-recorder/pull/177#issuecomment-1181996948)? Maybe some usage examples? I recall I was a bit confused on which part of the `ffmpeg` command output provided the exact string to be used with `wf-recorder` (maybe it was me being a bit thick)

llyyr:
Better version of your comment is already in the readme after this PR. I don't mind doing more examples but it should be fairly obvious what wf-recorder can do. And if I start writing examples for ffmpeg functionality there won't be an end.

apiraino:
on a second thought: examples can also fit somewhere in the wiki.

ammen99 (approved):
We seem to have grown quite a long list of options. LGTM and thanks!

PR #181, opened 2022-07-26
Update calll to deprecated method

WhyNotHugo:


src/frame-writer.cpp
```diff
@@ -654,8 +654,7 @@ bool FrameWriter::add_frame(const uint8_t* pixels, int64_t usec, bool y_invert)
         filtered_frame->pict_type = AV_PICTURE_TYPE_NONE;
 
         // So we have a frame. Encode it!
-        AVPacket pkt;
-        av_init_packet(&pkt);
+        AVPacket pkt = *av_packet_alloc();
```
ammen99, line 657: This seems as if we are leaking the packet. Isn't that true?

WhyNotHugo:
I believe you are correct, I will review this.

ammen99:
The same changes have been proposed and also merged in #193, but thank you for the PR nonetheless :)

WhyNotHugo:
Apologies for never getting back.

Glad that this has made it in anyway, cheers!

PR #182, opened 2022-07-27
fix buffersink pix_fmt handling

llyyr:
There was a bug caused by 5384bc9f3691833b5bf63e97afb953eadfbbec90 which
caused wf-recorder to select incorrect pix_fmt for buffersink, and
to ignore -x/--pixel_format. This is fixed by using the internal
choose_sw_format() function again.

Also I'm unsure what the point of --force-yuv being available was when the user can just set it with --pixel-format anyway. Shouldn't this just be removed in favor of suggesting "--pixel-format yuv420p" instead?

### DIFF
--- a/manpage/wf-recorder.1
+++ b/manpage/wf-recorder.1
@@ -1,4 +1,4 @@
-.Dd $Mdocdate: July 26 2022 $
+.Dd $Mdocdate: July 30 2022 $
 .Dt WF-RECORDER 1
 .Os
 .Sh NAME
@@ -6,7 +6,7 @@
 .Nd simple screen recording program for wlroots-based compositors
 .Sh SYNOPSIS
 .Nm wf-recorder
-.Op Fl abcCdDefFghlmopPrRtvxX
+.Op Fl abcCdDefFghlmopPrRvxX
 .Op Fl a, -audio Op Ar =DEVICE
 .Op Fl b, -bframes Ar max_b_frames
 .Op Fl c, -codec Ar output_codec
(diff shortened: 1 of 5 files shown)


llyyr:
Removed the `--force-yuv` option as discussed on IRC

llyyr:
@ammen99 can you take a look at this again?

manpage/wf-recorder.1
```diff
@@ -206,7 +201,7 @@ data for
 .Ql vaapi
 input but really only support yuv planar formats.
 In this case, use the
-.Fl t , -force-yuv
+.Fl x yuv420p , -pixel-format yuv420p
```
ammen99, line 204: Something is wrong here, after rendering it looks like this:

```
In this case, use the -x -yuv420p,
     --pixel-format -yuv420p option in addition to the ‘vaapi’ options to convert the data to yuv planar data before sending it to the GPU.
```

llyyr: Hm it renders fine for me but I changed it anyway since it was redundant

ammen99: It is better now, but I still get `use the -x -yuv420p option` with both neovim and the default manpager.

ammen99: In fact, some other options (which you have not changed at all) are also broken in the manpage, I wonder what is going on.

ammen99: Aha, so you need `.Fl x Ar yuv420p`

llyyr: I changed everything that seemed like obvious inconsistencies to me

ammen99 (approved):
Thanks!

PR #184, opened 2022-09-07
Add support for 16 bit deep formats

### LINKED ISSUE #155: Support for high bit depth screencopy buffers

mstoeckl:
This should be enough to close #155 . 

I'm making this PR very early, because I sometimes use wf-recorder to test that sway/wlroots with high bit depth work properly. It may be quite a while before the prerequisites to actually use the new formats from this PR will be available in released software.

This can be tested with FFmpeg 6.0 (released early 2023), and sway built from the [mstoeckl/sway/extra-bit-depths](https://github.com/mstoeckl/sway/commits/extra-bit-depths) branch, and wlroots using the Vulkan renderer (requires git master or >= 0.17.0, set `WLR_RENDERER=vulkan`). Run a nested instance of sway, and invoke `swaymsg output '*' render_bit_depth 16` or `swaymsg output '*' render_bit_depth 16f` to set the output buffer bit depth. (A nested instance is recommended here, because to test this with sway with the DRM backend, you'd need a relatively recent GPU so that the system that lists the appropriate plane formats in `drm_info`.)


ammen99:
> This can be tested with FFmpeg 6.0 (released early 2023)

Does this mean it won't compile with some earlier versions? Or will it just fail at runtime? It seems at least to compile here on FFmpeg 5.1.2.

ammen99:
Ok, then we can safely merge. Thanks!

ammen99 (approved):

PR #190, opened 2022-10-28
FrameRate Changes, Implement CFR and estimated buffer framerate

Quackdoc:
this sets the framerate toggle to a proper CFR, and implements buffrate for encoder hinting

buffrate is used to preserve VFR on encoders that require a framerate cap or target framerate (IE encoders without proper VFR support). it can also be used as a soft fps cap, but unless set at an integer value of the display, it may provide poor results.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -1,7 +1,6 @@
 #pragma once
 
 #define DEFAULT_CODEC "@default_codec@"
-#define DEFAULT_FRAMERATE @default_framerate@
 #define DEFAULT_AUDIO_CODEC "@default_audio_codec@"
 #define DEFAULT_AUDIO_SAMPLE_RATE @default_audio_sample_rate@
 #define FALLBACK_AUDIO_SAMPLE_FMT "@fallback_audio_sample_fmt@"
--- a/manpage/wf-recorder.1
+++ b/manpage/wf-recorder.1
@@ -69,7 +69,7 @@ To modify codec parameters, use
 .Fl p Ar option_name=option_value
 .Pp
 .It Fl r , -framerate Ar framerate
-Changes an approximation of the video framerate. The default is 60.
+Changes an approximation of the video framerate.
 .Pp
 .It Fl d , -device Ar encoding_device
 Selects the device to use when encoding the video.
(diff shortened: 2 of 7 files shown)


ammen99:
Hi, thank you for working on this, but I think the docs need to be clarified. Even I am not sure what buffrate is, and what framerate is (basically, what is the difference between -B and -r)? When would I use the one, when the other?

Quackdoc:
Ill amend this to be clear when I get the chance, but buffrate sets the buffer input device's framerate, where as framerate uses the fps frame filter to set framerate, 

the reason for the distinction is that when we set the buffer device framerate, you use the video buffer video filter, which in ffmpeg more or less acts as a video source device. so what we are doing is telling ffmpeg the framerate to expect as the input stream. this is not a hard framerate, but rather a maximum framerate to expect, this lets it work as a framerate cap so to speak. this means that ffmpeg should never encode more FPS then the buffrate, it also however does still allow sending lower FPS then the buffer is set to, hence preserving VFR while setting a cap on the framerate.

this is necessary since there are numerous encoders and I think even a muxer or two that need this information to work properly. (in my testing I primarily used SVT-AV1).

Framerate on the other hand uses the ffmpeg fps filter to set a constant framerate, this works in most cases by duplicating frames until the fps target is hit, this is necessary for a lot of users as VFR is relatively unsupported in a lot of applications. (some video editors, and even some video sharing applications do not work with VFR videos)

I originally thought to leave Framerate to mean the buffer rate, but I think most users would expect framerate to be equal to CFR as this is what most other software does (from video editors, to handbrake). although this could be done by the user manually specifing the fps filter, I believe it would be best if this had its own toggle for the user's experience.

I also thought about renaming buffrate to max FPS to hint towards the implicated use of using it to preserve VFR. however I am not sure that all encoders will play nice with that.

ammen99 (approved):
LGTM, thanks for the clarification!

Quackdoc:
Sorry for the super late update, I won't lie I completely forgot about this. I changed the docs to be more clear, also instead of suggesting that it might work as a softcap for fps. that seems rather inconsistent so I left it out

PR #193, opened 2022-11-26
Fix deprecation warnings

llyyr:


ammen99 (approved):
LGTM, thanks!

WhyNotHugo:
This breaks building from tarballs, since git metadata is absent. Maybe this should be skipped if `.git` does not exist?

PR #197, opened 2022-12-18
Add default codec parameters for libvpx/libvpx-vp9.

alebastr:
Just a minimal set of options to enable realtime encoding.

I have no idea what rate control options should be used, so I'm omitting these. Feel free to suggest a good defaults.

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -44,19 +44,35 @@ void FrameWriter::init_hw_accel()
 
 void FrameWriter::load_codec_options(AVDictionary **dict)
 {
-    static const std::map<std::string, std::string> default_x264_options = {
-         {"tune", "zerolatency"},
-         {"preset", "ultrafast"},
-         {"crf", "20"},
+    using CodecOptions = std::map<std::string, std::string>;
+
+    static const CodecOptions default_x264_options = {
+        {"tune", "zerolatency"},
+        {"preset", "ultrafast"},
+        {"crf", "20"},
     };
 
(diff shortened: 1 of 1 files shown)


src/frame-writer.cpp
```diff
@@ -44,19 +44,35 @@ void FrameWriter::init_hw_accel()
...
+        {"cpu-used", "5"},
+        {"deadline", "realtime"},
     };
 
-    if (params.codec.find("libx264") != std::string::npos ||
-        params.codec.find("libx265") != std::string::npos)
+    static const std::map<std::string, const CodecOptions&> default_codec_options = {
+        {"libx264", default_x264_options},
+        {"libx265", default_x264_options},
+        {"libvpx", default_libvpx_options},
```
Conan-Kudo, line 63: Shouldn't there be also a line for `libvpx-vp9`?

Conan-Kudo (changes requested):

src/frame-writer.cpp
alebastr: No, it's a substring search so `libvpx` also matches `libvpx-vp9`.
The original code did that to avoid listing codec variants like `libx264rgb` and I preserved the behavior.

Conan-Kudo: Ah, makes sense. Thanks for explaining.

Conan-Kudo (approved):

ammen99 (approved):
Makes sense, thanks!

PR #198, opened 2022-12-21
Use VP9+Opus on MKV by default

### LINKED ISSUE #196: libvpx fails to run
I'm trying to re-package wf-recorder for fedora now that fedora has separated ffmpeg and ffmpeg-free. I'm trying to find a codec that will work with both. Obviously the default libx264 is not available in ffmpeg-free and libopenx264 is not in ffmpeg. 

I've been trying libvpx-v9 but wf-recorder takes >25s to respond to SIGINT (^C) - most users would give up, they would use pkill (SIGTERM) and find a corrupt mp4 file.

libvpx fails with ffmpeg or ffmpeg-free:
```
[mp4 @ 0x7fc058000d40] Could not find tag for codec vp8 in stream #0, codec not currently supported in container
```

I guess this is the bug that I'd like to report here but also ask what your suggestion would be for a universal codec. I should probably stay away from *vaapi|qsv|v4l2m2m|amf *  as these have hardware dependencies. *jpeg|webp* are probably not useful in this context. 

I'm told that theora is outdated and av1 is slow. Any other suggestions?

Conan-Kudo:
This PR stacks on top of #197 to change the default for wf-recorder to use only royalty-free codecs by default.

Fixes #196

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,12 +63,12 @@ git clone https://github.com/ammen99/wf-recorder.git && cd wf-recorder
 meson build --prefix=/usr --buildtype=release
 ninja -C build
 ```
-Optionally configure with `-Ddefault_codec='codec'`. The default is libx264. Now you can just run `./build/wf-recorder` or install it with `sudo ninja -C build install`.
+Optionally configure with `-Ddefault_codec='codec'`. The default is libvpx-vp9. Now you can just run `./build/wf-recorder` or install it with `sudo ninja -C build install`.
 
 The man page can be read with `man ./manpage/wf-recorder.1`.
 
 # Usage
-In its simplest form, run `wf-recorder` to start recording and use Ctrl+C to stop. This will create a file called `recording.mp4` in the current working directory using the default codec.
+In its simplest form, run `wf-recorder` to start recording and use Ctrl+C to stop. This will create a file called `recording.mkv` in the current working directory using the default codec.
 
 Use `-f <filename>` to specify the output file. In case of multiple outputs, you'll first be prompted to select the output you want to record. If you know the output name beforehand, you can use the `-o <output name>` option.
 
@@ -81,7 +81,7 @@ wf-recorder -g "$(slurp)"
 You can record screen and sound simultaneously with
 
 ```
-wf-recorder --audio --file=recording_with_audio.mp4
+wf-recorder --audio --file=recording_with_audio.mkv
 ```
 
 To specify an audio device, use the `-a<device>` or `--audio=<device>` options.
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ xbps-install -S wf-recorder
 
 ## Fedora Linux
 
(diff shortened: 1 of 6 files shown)


Conan-Kudo:
@alebastr, @bhepple: this might help for #196 and the Fedora packaging of `wf-recorder`.

Conan-Kudo:
@bhepple confirmed for me this works as expected, so I'm opening this for review.

ammen99:
It seems fine to me to add the default paramts for vp9, but why does this have to be changed upstream? Can't fedora packagers override the defaults? That's why they are defined as meson options.

Conan-Kudo:
> It seems fine to me to add the default paramts for vp9, but why does this have to be changed upstream? 

Default params are in #197

> Can't fedora packagers override the defaults? That's why they are defined as meson options.

I can certainly make the container default configurable too. Though, my plea is that the upstream default should use royalty free codecs because it's better for the user, distributor, and the ecosystem at large. We should promote formats that aren't encumbered and cause problems for people depending on where they are.

ammen99 (approved):

ammen99:
Thanks!

llyyr:
I'm not sure this is a good change, most users expect their videos to work on the web but most mobile apps and even a lot of instant messaging apps such as Facebook/Whatsapp/Line/Discord/Twitter only support h264. It's fine to add sensible defaults for libvpx but I don't agree with the change that wf-recorder should default to using VP9+Opus and the mkv container.

Conan-Kudo:
> Someone who's not familiar with video codecs uploading their video on Facebook and finding out that it doesn't work for some mysterious reason is going to cause more issues than someone using a specific Linux distro that doesn't happen to have hardware decoding for a certain video codec.

So I just did some testing the services I do have:

- Telegram supports VP8 WebM
- Discord supports VP8 WebM
- Element supports VP8 WebM
- Facebook supports VP8 WebM and VP9 MKV
- Twitter doesn't even *let* me upload *any* video.

I can change the default to VP8+Vorbis WebM, which works across all these services. It also matches the default for GNOME and other desktops that have built-in screen recording.

ammen99:
I think GNOME's default should be good enough for us.

0atman:
These changes just came through to arch and broke my workflow, which until today relied on the default being mp4.
(My editor, reaper.fm, which uses libffmpeg and libvlc can't read the files `wf-recorder` now produces by default.)

It's no trouble for me to look up the right combination of command flags to switch it back to something that I can use, but please note that changing the default behaviour has huge effects on your users that we do not expect from a minor version release.

0atman:
I'm not sure I agree with @Conan-Kudo 's reasoning that if all these social media platforms support webm, that makes it good. Who screenrecords and upload without at least TRIMMING the start and end of the video? 

You should have considered what your users are actually doing with the files. Are they uploading them as-is, or are they editing them?

ammen99:
@0atman The change was later reverted, the default in 0.4 is libx264 again. Though downstream distros might have changed the defaults (though that isn't something I can influence).

PR #199, opened 2022-12-23
meson: Use gnu variants of C/C++ standards

Conan-Kudo:
This fixes the build for POWER systems because GNU variants force boolean types to follow the C/C++ standard (and remain scalar) rather than becoming vector with the POWER AltiVec extensions.

This has no change in behavior for non-POWER architectures.

Reference: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58241

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -6,8 +6,8 @@ project(
 	license: 'MIT',
 	meson_version: '>=0.47.0',
 	default_options: [
-		'cpp_std=c++11',
-        'c_std=c11',
+		'cpp_std=gnu++11',
+		'c_std=gnu11',
 		'warning_level=2',
 		'werror=false',
 	],


ammen99:
I am not sure why this should be upstream. The project uses the C++11 standard. Note that users/packagers can override these options on the command line by simply using `-Dcpp_std=gnu++11`. Isn't that enough?

Conan-Kudo:
The reason is that using GNU++11 ensures that you're following the C++11 standard on architectures where the compiler's C++11 mode is ambiguous. While it is true we can just pass the mode, in, I figure it makes sense to have sane defaults here so nobody has to try to figure out how to fix it when they run into this issue when building for POWER systems.

ammen99:
As discussed on IRC, I don't like the idea of merging workarounds for specific architectures (when there is a simple fix during the build process), so NACK for this change.

Thanks for investigating the issue nonetheless, I guess other people may find the information useful.

PR #201, opened 2022-12-24
Add GitHub Actions CI to test the build

Conan-Kudo:
This will hopefully ensure that future pull requests get basic compilation testing to prevent broken merges.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,30 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+
+    container: registry.fedoraproject.org/fedora:latest
+
+    steps:
+      - name: Set up DNF download cache
+        id: dnf-cache
+        uses: actions/cache@v3
+        with:
+          path: /var/cache/dnf
+          key: ${{ runner.os }}-dnfcache
+      - name: Install pre-requisites
+        run: dnf --assumeyes --setopt=install_weak_deps=False install
+              gcc-c++ meson /usr/bin/git /usr/bin/wayland-scanner
+              'pkgconfig(wayland-client)' 'pkgconfig(wayland-protocols)' 'pkgconfig(libpulse-simple)'
+              'pkgconfig(libavutil)' 'pkgconfig(libavcodec)' 'pkgconfig(libavformat)'
+              'pkgconfig(libavdevice)' 'pkgconfig(libavfilter)' 'pkgconfig(libswresample)'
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
I don't know what's going on here, but it definitely runs in my branch: https://github.com/Conan-Kudo/wf-recorder/actions/runs/3771840054/jobs/6412444975

Conan-Kudo:
It's probably not showing up on the pull request here because actions might not be enabled in this repo yet?

README.md
```diff
@@ -54,7 +54,7 @@ sudo apt install libavutil-dev libavcodec-dev libavformat-dev libswscale-dev lib
 
 #### Fedora
 ```
-$ sudo dnf install wayland-devel wayland-protocols-devel ffmpeg-free-devel
+$ sudo dnf install wayland-devel wayland-protocols-devel ffmpeg-free-devel pulseaudio-libs-devel
```
ammen99, line 57: Seems like this doesn't belong in this commit.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,30 @@
...
+              'pkgconfig(wayland-client)' 'pkgconfig(wayland-protocols)' 'pkgconfig(libpulse-simple)'
+              'pkgconfig(libavutil)' 'pkgconfig(libavcodec)' 'pkgconfig(libavformat)'
+              'pkgconfig(libavdevice)' 'pkgconfig(libavfilter)' 'pkgconfig(libswresample)'
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # Shallow clones speed things up
+      - name: meson configure
+        run: meson ./Build
+      - name: compile with ninja
+        run: ninja -C ./Build -j4
```
ammen99, line 30: Does GitHub have more than 1 build job? ninja usually should autodetect the best amount of threads to use.

Conan-Kudo: I usually set it to a specific number of cores to make it predictable, but I can drop it to make it dynamic...

ammen99: I think hardcoding 4 threads doesn't make sense for a GitHub Action. AFAIK they are single threaded anyway ..

README.md
Conan-Kudo: Dropped and moved to #203

.github/workflows/build.yaml
Conan-Kudo: Dropped.

ammen99:
Let's merge and see whether the Github Action will be enabled automatically (I think it should be). In any case, thanks, this is a very useful addition :)

PR #203, opened 2022-12-25
README: Flesh out build dependencies for building from source

Conan-Kudo:
When building on Ubuntu or Fedora, you need Meson and a C++ compiler. Additionally, add pulseaudio-libs-devel to have audio functionality for Fedora.


ammen99 (approved):
Thanks!

PR #209, opened 2023-01-17
Only flush when video codec has delay

nowrep:
Fixes error on ending recording.

Application provided invalid, non monotonically
increasing dts to muxer in stream 0: 58995 >= 56744


ammen99 (approved):
LGTM, thanks!

PR #215, opened 2023-03-11
Use codec format which best matches input

mstoeckl:
This change will, if the codec cannot directly handle the input format, make wf-recorder choose the best matching format according to the ranking provided by `av_find_best_pix_fmt_of_2`,  instead of falling back to yuv420p or the first listed codec format.

Note that wf-recorder currently defaults to the libvpx (VP8) recorder, which only supports (yuv420p yuva420p). For testing, I'd recommend using either `--codec libx264`  or `--codec libvpx-vp9`, since both of those support a much wider range of input formats.


ammen99 (approved):
Didn't know about the existence of this function, thanks!

Issue #214, opened 2023-03-06
Default poor libvpx quality (master branch)


apiraino:
So, now wf-recorder defaults to the VP8 encoder instead of h264.

I'm running wf-recorder compiled from 460d454 and I observe that the defaults for the libvpx codec produce a low quality ("blockier") video while defaults for h264 produced a better video, example:
```
wf-recorder -a --log
```

The low quality of the video is also explained by this warning
```
[libvpx @ 0x7fbd240022c0] Neither bitrate nor constrained quality specified, using default CRF of 32 and bitrate of 256kbit/sec
``` 

As per [this issue](https://github.com/ammen99/wf-recorder/issues/79#issuecomment-615412987), I tried to tweak the quality with this command but unfortunately the previous warning persists:
```
wf-recorder -a -pqmin=0 -pqmax={tried from 32 to 50} --log
```

How can this codec be configured for better quality?

cc @Conan-Kudo that authored #198 and hopefully more knowledgeable. Can you please share a configuration for the new default codec that results in a good video quality? 

thanks

ammen99:
As an alternative, you can always choose another codec with -c (vp9, libx264 are probably better alternatives if your ffmpeg and video players support them, which is usually the case)

apiraino:
@ammen99 thanks for the tip. The thing is that tweaking [VP8](https://trac.ffmpeg.org/wiki/Encode/VP8) or [VP9](https://trac.ffmpeg.org/wiki/Encode/VP9#speed) is tricky, I'm trying to force wf-recorder to use the parameters I pass from the command line, no luck for now. Examples:
```
# error
wf-recorder -a -c libvpx-vp9 -pminrate=500K -pmaxrate=500K -pb:v=500K
# params ignored?
wf-recorder -a -c libvpx-vp9 -pcrf=30 -pb:v=0
# params ignored?
wf-recorder -a -c libvpx-vp9 -pcrf=30 -pb:v=2000K
```

In addition VP9 needs some additional params (`-pdeadline=realtime -pspeed=8`) to reduce the lag when the encoder is crunching frames (also mentioned in #80 and [here](https://github.com/ammen99/wf-recorder/issues/196#issuecomment-1356618229)).

I should probably also mention that previously wf-recorder (using the default h264 codec) produced the following file:
```
$ wf-recorder -a -c libx264
...
Setting codec option: crf=20
Setting codec option: preset=ultrafast
Setting codec option: tune=zerolatency

# Bitrate video
[...]

PR #216, opened 2023-03-12
Improve quality and performance of WebM-based recordings

Conan-Kudo:
Suggested-by: @apiraino

Fixes: #214

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ git clone https://github.com/ammen99/wf-recorder.git && cd wf-recorder
 meson build --prefix=/usr --buildtype=release
 ninja -C build
 ```
-Optionally configure with `-Ddefault_codec='codec'`. The default is libvpx. Now you can just run `./build/wf-recorder` or install it with `sudo ninja -C build install`.
+Optionally configure with `-Ddefault_codec='codec'`. The default is libvpx-vp9. Now you can just run `./build/wf-recorder` or install it with `sudo ninja -C build install`.
 
 The man page can be read with `man ./manpage/wf-recorder.1`.
 
@@ -108,8 +108,8 @@ To set a specific output format, use the `--muxer` option. For example, to outpu
 wf-recorder --muxer=v4l2 --codec=rawvideo --file=/dev/video2
 ```
 
-To use GPU encoding, use a VAAPI codec (for ex. `vp8_vaapi`) and specify a GPU device to use with the `-d` option:
+To use GPU encoding, use a VAAPI codec (for ex. `vp9_vaapi`) and specify a GPU device to use with the `-d` option:
 ```
-wf-recorder -f test-vaapi.webm -c vp8_vaapi -d /dev/dri/renderD128
+wf-recorder -f test-vaapi.webm -c vp9_vaapi -d /dev/dri/renderD128
 ```
 Some drivers report support for rgb0 data for vaapi input but really only support yuv planar formats. In this case, use the `-x yuv420p` or `--pixel-format yuv420p` option in addition to the vaapi options to convert the data to yuv planar data before sending it to the GPU.
--- a/manpage/wf-recorder.1
+++ b/manpage/wf-recorder.1
@@ -190,12 +190,12 @@ loopback you might use:
 .Dl $ wf-recorder --muxer=v4l2 --codec=rawvideo --file=/dev/video2
 .Pp
 To use GPU encoding, use a VAAPI codec (for ex.
-.Ql vp8_vaapi
+.Ql vp9_vaapi
 ) and specify a GPU
 device to use with the
 .Fl d
 option:
-.Dl $ wf-recorder -f test-vaapi.webm -c vp8_vaapi -d /dev/dri/renderD128
+.Dl $ wf-recorder -f test-vaapi.webm -c vp9_vaapi -d /dev/dri/renderD128
 .Pp
 Some drivers report support for
 .Ql rgb0
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,4 +1,4 @@
-option('default_codec', type: 'string', value: 'libvpx', description: 'Codec that will be used by default')
+option('default_codec', type: 'string', value: 'libvpx-vp9', description: 'Codec that will be used by default')
 option('default_audio_codec', type: 'string', value: 'libvorbis', description: 'Audio codec that will be used by default')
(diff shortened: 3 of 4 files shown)


Quackdoc:
I dont think VP9 is a good idea, it's pretty slow. especially at cpu-used 5, even with cpu-used 8, my chuwi hi10x couldn't even begin to think about recording realtime 1080p30 let alone 60. even my r5 2600 struggles at encoding vp9 realtime with plenty of tuning

ammen99:
> I dont think VP9 is a good idea, it's pretty slow. especially at cpu-used 5, even with cpu-used 8, my chuwi hi10x couldn't even begin to think about recording realtime 1080p30 let alone 60. even my r5 2600 struggles at encoding vp9 realtime with plenty of tuning

How does that compare to vp8 and libx264 (the older default)?

llyyr:
With default settings, my i5-13600k encodes with libx264 at 13x speed, vp8 at 1.3-1.5x speed (it varies a lot), and vp9 at around 0.8-0.85x speed.

I'm still in favor of defaulting to libx264 because the benefits it brings over vpx are just too big, both in terms of compatibility over the web as well as hardware decoding capabilities. And as mentioned in #215, vp8 only supports yuv420p pixel format.

I don't know why Fedora can't just set vp8 or vp9 as the default codec during build, if that's what they want. To be clear, I can also just set libx264 as the default when building but using vpx as the upstream defaults inconveniences more people while only serving the purposes of one distro.

Quackdoc:
> How does that compare to vp8 and libx264 (the older default)?

on the chuwi tablet, libx264 just manages to just stably encode x264 1080p30 when nothing else is going on, libvpx does about 20fps, vp9 does 8fps.

at 720p vp8 does about 29-31 fps, vp9 does about 6fps, libx264 does about 40fps

on my 2600 at 1080p libvpx does 100fps, 1080p30 libvpx-vp9 does about 70 when nothing else is going on, and libx264 does about 150fps~

all of the tests for vp8/vp9 were tested with row-mt which increases speed.

Conan-Kudo:
> using vpx as the upstream defaults inconveniences more people while only serving the purposes of one distro.

Aside from the fact that it's not just Fedora that can't ship libx264 (Fedora, openSUSE, and their downstreams can't), using a codec that's encumbered by default is (IMO) ethically wrong.

To be honest, I would prefer to use AV1, but that's even _more_ complex. Most of my experience has been with the AOM encoder, which is abysmally slow. [The most performant CPU encoder is SVT-AV1](https://medium.com/@ewoutterhoeven/av1-is-ready-for-prime-time-svt-av1-beats-x265-and-libvpx-in-quality-bitrate-and-speed-31c1960703db), but I haven't had a chance to test it on my older machines to see if it's good for this sort of thing.

(Honestly, in an ideal world, wf-recorder would probe for hardware accelerated variants of codecs and use them if they work, but because ffmpeg is dumb about how it exposes them, I get why it doesn't.)

apiraino:
It looks like that different people have different (and valid) opinions on what the default encoding codec should be.  How hard would it be to make it configurable without recompiling? :smiley: 

`ffmpeg` infers the encoding codec based on the file extension choosen by the user.

ammen99:
> It looks like that different people have different (and valid) opinions on what the default encoding codec should be. How hard would it be to make it configurable without recompiling? 😃
> 
> `ffmpeg` infers the encoding codec based on the file extension choosen by the user.

It has been possible to choose the codec at runtime since basically forever ;) See `-c`

apiraino:
> It has been possible to choose the codec at runtime since basically forever ;) See `-c`

Sure and it's very handy, I use it to create different profiles. Would that also work for you @Quackdoc - if you need to stick to x264?

(my understanding is that the discussion here revolves around the _default hardcoded_ codec)

llyyr:
In my opinion, the default should be what is expected by most users, and if you're going to change it then there should be a very good reason for it. wf-recorder had libx264 as the default for a very long time, and it is also what is expected (for example, simplescreenrecorder also defaults to x264).

If you're telling the majority of your users to use a flag to get the behavior that they want, then doesn't that kind of defeat the purpose of this PR? I could also tell you to go use `-p/--codec-param` if I were to take that logic to its logical conclusion.

not to mention we're now recommending people to use vp{8,9}_vaapi, which only works on GTX 900 series GPUs or newer for vp8 and GTX 1000 series for vp9. And AMD doesn't support vp8 or vp9 hardware encoding at all. Promoting free software is fine and all, but I think you're losing the plot a bit when your ethical pursuit leads you to recommending defaults that only work on NoVideo GPUs.

Conan-Kudo:
> I think you're losing the plot a bit when your ethical pursuit leads you to recommending defaults that only work on NoVideo GPUs.

It also works on nearly every Intel CPU since 2016.

llyyr:
Why should we be recommending params that's only going to work on relatively new hardware by certain vendors? The defaults should be what's safest and what's most likely to work in the most types of environments. Makes no sense to suggest the defaults that's more likely to not work than it is to work.

Quackdoc:
> Sure and it's very handy, I use it to create different profiles. Would that also work for you @Quackdoc - if you need to stick to x264?

I'm fine with using custom settings, I already have a bunch of alias' setup anyways, I just don't think using a codec thats going to be effectively unusable as a default is a good idea. vp9 will straight up not work on plenty of hardware no matter how much you tune it sadly. and since ffmpeg doesn't support svt-vp9 I doubt that's going to change anytime soon. 

for software:
Ideally default would be AVC since that by far gives the best quality:preformance at a reasonable filesize. VP8 is a "reliable" fallback, however quality of it will always be rather poor. VP9 does look significantly better but at the expense that only moderate to high end computers can *reliably* record it. since you either need enough per core perf to brute force encoding it and doing other tasks, or have enough cores that libvpx-vp9 can run comfortably without being interfered with. (which even then isnt that easy without tweaking).

as for libsvtav1, it is marginally faster then libvpx-vp9 but still too significantly slower then vp8, especially since those speed benefits come from better utilization of available resources, which said resources are also being used by the system, so you are much to the same issue it's hard to do even remotely *reliably*

as for hwdec, can't we just say "use `vp9_vaapi` or `h264_vaapi`"? I don't think it really matters what the example uses as long as the error it gives if it's missing is fine.

ammen99:
Seems to me we should go back to libx264. I get that its not available everywhere, but 1) unfounded intuition tells me most users would have that codec installed 2) it seems to be the best wrt speed/quality, and that's what we want. I'll send a PR shortly.

PR #219, opened 2023-08-05
Reduce memory usage with DMA-BUF capture

nowrep:
Don't allocate extra buffer when mapping DRM frame. Also keep the initial pool size default.


ammen99 (approved):
Seems meaningful, even though I am no ffmpeg expert ..
Thanks :)

PR #220, opened 2023-08-06
Use gbm_bo_create_with_modifiers + don't encode if capture fails

nowrep:


src/main.cpp
```diff
@@ -317,8 +317,14 @@ static void frame_handle_linux_dmabuf(void *, struct zwlr_screencopy_frame_v1 *f
     buffer.height = height;
 
     if (!buffer.wl_buffer) {
-        buffer.bo = gbm_bo_create(gbm_device, buffer.width,
-            buffer.height, format, GBM_BO_USE_LINEAR | GBM_BO_USE_RENDERING);
+        const uint64_t modifier = 0; // DRM_FORMAT_MOD_LINEAR
+        buffer.bo = gbm_bo_create_with_modifiers(gbm_device, buffer.width,
```
ammen99, line 321: Would you mind explaining why we try `with_modifiers` first, when should one use the two functions for allocation here?

nowrep: `gbm_bo_create_with_modifiers` is always preferable when supported (kernel supports modifiers). There are however still some GPUs without modifiers support, and with those the `gbm_bo_create_with_modifiers` will fail and we have to fallback to `gbm_bo_create`.

`gbm_bo_create` should work everywhere, but it will return `DRM_FORMAT_MOD_INVALID` which may be problematic - for example Vulkan doesn't allow importing dmabufs with invalid modifier.

ammen99 (approved):
LGTM, thanks!

src/main.cpp
nowrep: I don't know of any documentation explaining this, sorry. Most of my understanding comes from studying code.

Similar logic in wlroots https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/render/allocator/gbm.c#L92

Actually the dmabuf screencopy code is based on wlroots example which uses `gbm_bo_create` only and up until now I didn't know it will always return invalid modifier. For some reason I expected it to return some (best?) modifier if kernel supports modifiers, and I didn't confirm it because it seemed to just work. Also gbm documentation doesn't  mention it either.

PR #221, opened 2023-08-09
Dynamically increase number of used buffers

nowrep:
Start with 2 buffers and increase it when encoding starts to fall behind capture.

### DIFF
--- a/src/buffer-pool.hpp
+++ b/src/buffer-pool.hpp
@@ -1,0 +1,78 @@
+#pragma once
+
+#include <array>
+#include <mutex>
+#include <atomic>
+#include <type_traits>
+
+class buffer_pool_buf
+{
+public:
+    bool ready_capture() const
+    {
+        return released;
+    }
+
+    bool ready_encode() const
+    {
+        return available;
+    }
+
+    std::atomic<bool> released{true}; // if the buffer can be used to store new pending frames
+    std::atomic<bool> available{false}; // if the buffer can be used to feed the encoder
+};
+
+template <class T, int N>
+class buffer_pool
+{
+public:
+    static_assert(std::is_base_of<buffer_pool_buf, T>::value, "T must be subclass of buffer_pool_buf");
+
+    size_t size() const
+    {
+        return N;
+    }
+
+    const T& at(size_t i) const
+    {
+        return bufs[i];
+    }
+
+    T& capture()
+    {
+        return bufs[capture_idx];
+    }
+
+    T& encode()
+    {
+        return bufs[encode_idx];
+    }
+
+    // Signal that the current capture buffer has been successfully obtained
+    // from the compositor and select the next buffer to capture in.
+    T& next_capture()
+    {
+        std::lock_guard<std::mutex> lock(mutex);
+        bufs[capture_idx].released = false;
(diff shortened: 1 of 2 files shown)


nowrep:
The major source of encode thread lag behind capture thread is the encoder initialization that happens after capturing first frame. We could discard first captured frame and wait for encoder to initialize to reduce the number of used buffers even further.

On my system with this change:

```
libx264 = needs around 5-7 buffers
vaapi   = needs 3-4 buffers
```

ammen99:
At this point, seems to me we should separate buffer management into a separate data structure (preferably also a different file) to keep the code readable.

src/buffer-pool.hpp
```diff
@@ -0,0 +1,77 @@
...
+        return bufs[encode_idx];
+    }
+
+    T& next_capture()
+    {
+        std::lock_guard<std::mutex> lock(mutex);
+        bufs[capture_idx].released = false;
+        bufs[capture_idx].available = true;
+        capture_idx = (capture_idx + 1) % bufs_size;
+        if (capture_idx == 0 && !bufs[0].ready_capture() && bufs_size < N)
```
ammen99, line 54: Since we basically have a circular buffer, I wonder why we need to enlarge only when `capture_idx==0`. Is it not possible that the buffer is full at some other point in time? For example the following scenario (with say 2 buffers):

- Capture frame 0
- Capture frame 1
- Encode frame 0
- Capture frame 0 -> capture_idx is incremented to 1, but it is not yet ready for capture -> what do we do?

src/buffer-pool.hpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    std::atomic<bool> released{true}; // if the buffer can be used to store new pending frames
+    std::atomic<bool> available{false}; // if the buffer can be used to feed the encoder
+};
+
+template <class T, int N>
+class buffer_pool
+{
+public:
```
ammen99, line 27:
```suggestion
public:
  static_assert(std::is_base_of_v<buffer_pool_buf, T>);
```

src/buffer-pool.hpp
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return bufs[capture_idx];
+    }
+
+    T& encode()
+    {
+        return bufs[encode_idx];
+    }
+
+    T& next_capture()
```
ammen99, line 48:
```suggestion
    // Signal that the current capture buffer has been successfully obtained from the compositor and select the next buffer to capture in.
    T& next_capture()
```

src/buffer-pool.hpp
```diff
@@ -0,0 +1,77 @@
...
+        capture_idx = (capture_idx + 1) % bufs_size;
+        if (capture_idx == 0 && !bufs[0].ready_capture() && bufs_size < N)
+        {
+            capture_idx = bufs_size;
+            bufs_size++;
+        }
+        return bufs[capture_idx];
+    }
+
+    T& next_encode()
```
ammen99, line 62:
```suggestion
    // Signal that the encode buffer has been submitted for encoding and select the next buffer for encoding.
    T& next_encode()
```

ammen99 (approved):
LGTM now, thanks!

PR #224, opened 2023-08-09
Add X2RGB10 format mapping for DMA-BUFs

nowrep:
Requires https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230809124554.160462-2-nowrep@gmail.com/
And for mesa driver it also requires https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/24353

Tested with `WLR_RENDERER=vulkan sway` + `output * render_bit_depth 10`.


ammen99:
Which version of ffmpeg is required for this code to compile? Is this also a case of a format available since ages, but simply not supported in the ffmpeg backend?

ammen99 (approved):
I guess there is no reason not to merge this then ;)

PR #225, opened 2023-08-23
add --no-dmabuf option

ammen99:
This is needed in case that the DMABUF path does not work due to system bugs.


ammen99:
@llyyr This option should be enough for you I hope ;)

PR #226, opened 2023-08-23
Skip git-based version if building from tarballs

WhyNotHugo:
This allows downstream distributions to builds from tarballs which lack any git metadata.


ammen99:
I am not sure, isn't it enough to just set `check: false` in the git commands?

WhyNotHugo:
Setting `check: false` would work. I don't have much context as to why this was changed; it happened in https://github.com/ammen99/wf-recorder/pull/193

ammen99:
I'd prefer to set `check: false` then, we do a check whether the command succeeds anyway.

ammen99:
Please write out `check: false` explicitly :)
Otherwise lgtm.

WhyNotHugo:
Done

ammen99 (approved):
Thanks!

PR #233, opened 2023-09-08
Also set buffer.drm_format with shm capture

nowrep:
HWFrames context uses drm_format to set correct sw_format even when not using DMA-BUF capture.


ammen99 (approved):
Nice catch, thanks!

PR #234, opened 2023-09-12
Add command line arguments for enabling and disabling cursor capture.

Stronkkey:
Added 2 optional command line arguments.

-s, --show-mouse         Shows the mouse in the output file.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -51,7 +51,7 @@ static struct zxdg_output_manager_v1 *xdg_output_manager = NULL;
 static struct zwlr_screencopy_manager_v1 *screencopy_manager = NULL;
 static struct zwp_linux_dmabuf_v1 *dmabuf = NULL;
 static struct wl_drm *drm = NULL;
-void request_next_frame();
+void request_next_frame(bool enable_mouse);
 
 struct wf_recorder_output
 {
@@ -259,12 +259,12 @@ static void frame_handle_ready(void *, struct zwlr_screencopy_frame_v1 *,
 }
 
 static void frame_handle_failed(void *, struct zwlr_screencopy_frame_v1 *) {
-    std::cerr << "Failed to copy frame, retrying..." << std::endl;
+    std::cerr << "Failed to copy frame, retrying...\n";
     ++frame_failed_cnt;
-    request_next_frame();
+    request_next_frame(false);
     if (frame_failed_cnt > MAX_FRAME_FAILURES)
     {
-        std::cerr << "Failed to copy frame too many times, exiting!" << std::endl;
+        std::cerr << "Failed to copy frame too many times, exiting!\n";
         exit_main_loop = true;
     }
 }
@@ -780,6 +780,10 @@ Use Ctrl+C to stop.)");
   -m, --muxer               Set the output format to a specific muxer instead of detecting it
                             from the filename.
 
+  -s, --show-mouse          Shows the mouse in the output file. This is the default behavior.
+
+  -e, --hide-mouse          Hides the mouse in the output file.
+
   -x, --pixel-format        Set the output pixel format. These can be found by running:
                             ffmpeg -pix_fmts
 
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -956,6 +959,10 @@ int main(int argc, char *argv[])
                 params.muxer = optarg;
                 break;
 
+            case 's':
+                params.show_mouse = 1;
```
apiraino, line 963: hey @Stronkkey just to check if I understand this patch correctly: currently `wf-recorder` records the mouse cursor (at least, I see it).

After this patch there should be a new option to ... _disable_ the mouse cursor or _enable_ it?

I'm slightly confused by the option name and the effect that it has :)

thanks

ammen99 (changes requested):
I am not opposed to the idea, two main things:

- We should use bool instead of int32_t, it is a boolean option after all
- I would prefer not to break existing behavior. That is, make the cursor shown by default, and the option should 'hide' the cursor.

Stronkkey:
Added a new option (--hide-mouse -e) and the cursor is now shown by default in the output file

src/main.cpp
```diff
@@ -780,6 +780,10 @@ Use Ctrl+C to stop.)");
   -m, --muxer               Set the output format to a specific muxer instead of detecting it
                             from the filename.
 
+  -s, --show-mouse          Shows the mouse in the output file. This is the default behavior.
```
ammen99, line 783: I don't think we need both options: the way it is now, `-s` has no effect (because it is the default).

src/main.cpp
```diff
@@ -780,6 +780,10 @@ Use Ctrl+C to stop.)");
...
+
+  -e, --hide-mouse          Hides the mouse in the output file.
```
ammen99, line 785: I find `-e` a bit unintuitive, maybe we can make this a long option only?

ammen99 (commented):
Forgot to mention also that the manpage will need to be updated, but that can be left for the end once we have finalized the implementation.

src/main.cpp
Stronkkey: The e is for --hid(E)-mouse.

src/main.cpp
Stronkkey: Fixed now.

src/main.cpp
Stronkkey: This is done for consistency. If you can hide your mouse, you should also be able to show it.

ammen99:
My points remain: I would like only `--hide-mouse` as a long option, and the man page should be updated @Stronkkey

apiraino:
thank you for invoking my review @Stronkkey but I'm not a `wf-recorder` maintainer, just a happy user following its development :slightly_smiling_face: 

As I mentioned in this [comment](https://github.com/ammen99/wf-recorder/pull/234#discussion_r1323430897) and reiterated by @ammen99 in this [other comment](https://github.com/ammen99/wf-recorder/pull/234#discussion_r1324305166), it seems that **the default behaviour of wf-recorder is to show the mouse pointer**. As such I don't see a compelling reason to have both options:
```
-s, --show-mouse
--hide-mouse
```

The option to hide the mouse should be enough, IIUC. If a user wants to see the mouse cursor they will simply do nothing. I'd rather not change a default _implicit_ behaviour (show the mouse cursor) to an _explicit_ option, it's a breaking change that could surprise users.

I am also mildly against the short option `-e` because I feel it's not greatly self-explaining.

Other than that, this PR adds a feature I personally would not use but - hey - this is not my project so the maintainer will have the last word on accepting the patch or declining it :slightly_smiling_face: 

In any case: thanks for taking the time to contribute to wf-recorder.

Stronkkey:
@ammen99 Ok so now the option to hide the mouse is gone (+rebased commits).
@apiraino Oops didn't notice.

michaelskyba:
Any update on the review?

ammen99:
> @ammen99 Ok so now the option to hide the mouse is gone (+rebased commits).

I think something went wrong, because the option is still there when I go to the source file ..

Also, make sure to update the man page ;)

Stronkkey:
I'll pass on this on :sleeping:

PR #235, opened 2023-09-14
Use dmabuf feedback instead of wl-drm to get compositor device

nowrep:
wl-drm is a private protocol and while all compositors support it now, it may be removed in future.

This adds dependency on libdrm to be able to get the device path from dev_t returned from dmabuf feedback main_device.


ammen99 (approved):
Thanks!

Issue #218, opened 2023-07-27
Output recording plays locally but not in Firefox


buttke:
Hi, when I produce a recording with `wf-recorder`, the video plays just fine locally with `mpv`. When I upload that video to, say, Discord, and try to play it in the browser, a preview appears with the correct length and a thumbnail, but playing the video does not work.

https://github.com/ammen99/wf-recorder/assets/88161263/7726efc1-7366-4112-b44a-aa6f55f27389

buttke:
This issue occurs in Firefox but not Chromium.

buttke:
duplicate of #152 

The `-t` flag solves this. https://github.com/ammen99/wf-recorder/issues/152#issuecomment-1166593274

soreau:
With latest wf-recorder 0e7b2e27, do videos work in firefox/discord/messenger work with `wf-recorder -x yuv420p`? The default is yuv444p which I think may be the core problem.

Referencing related issues #189, #175, #152 and #28.

ItsDrike:
Hey, could you perhaps make a new release with this fix? Most distributions only ship releases, and this change was only made after the latest release (which was made Sep 5). If you have some pending changes, perhaps just cherry-pick this change and only put out a release with this single change, however it seems like a pretty big issue and it's weird that it's still not in any released version.

Issue #189, opened 2022-10-09
Cannot send videos recordered with wf-recorder through FB Messenger


cpouldev:
Hi, thanks for the app!

As the title says, I can't send any videos through FB Messenger (browser & android). Tried already on 2 different machines.

Strangely, the only videos I can send are the very short ones (eg 2-6 sec). For larger ones, facebook's dev console threw 500 http error on upload.php network call. You can (maybe?) replicate it by trying to send a 30s or 1m video via browser.

Screencasts from a Gnome session using its native recorder can be uploaded without any issues.

The command I use is this one

`wf-recorder -g "$(slurp)" -t  -f ~/Videos/screen_$(date +%Y-%m-%d_%H-%M-%S).mp4`

--

I checked past issues (https://github.com/ammen99/wf-recorder/issues/152) and the only thing that comes into my mind is the graceful termination of wf-recorder. Tried already the relevant PR but the issue still exists.

This is the only thing that keeps me back from using Sway as daily driver, any help will be deeply appreciated!

stefonarch:
Not sure if related, but kdenlive refuses to load recording.mp4 created by wf-recorder:
![screen_window_dom_12:17:41_](https://user-images.githubusercontent.com/10681413/194751215-21dbb6a4-8243-4154-92e0-83fa81ff7cc0.png)

ammen99:
If it is about variable refresh rate, try `--no-damage`: https://github.com/ammen99/wf-recorder/blob/master/manpage/wf-recorder.1#L85

stefonarch:
For kdenlive this makes no diff, same message.

cpouldev:
same for the browser. Can't play the video on android either. Here is a sample video file

https://user-images.githubusercontent.com/8311776/194762995-7b5c0198-7a1c-4e8b-911a-9e2e00a867da.mp4

soreau:
Perhaps this is the same as #175 and the -t option will help?

cpouldev:
I have the -t option already

Quackdoc:
@cpouldev if you can, try to build the #190 and see if it fixes your issues, I would recommend specifying -r 30/60 or whatever your desired framerate is.

cpouldev:
@Quackdoc thanks for your recommendation. I just built it and tried it but unfortunately the issue still persists.

If you have fb messenger, could you please try and send a full screen video (more than 40-50sec)? As I see it, the failing videos are bigger than 10MB and 30s+ long.

[...]

Issue #175, opened 2022-06-22
Default h264 videos do not play anymore in Firefox


apiraino:
When recording with `wf-recorder -f out.mp4` (leaving all defaults), the generated file does not play anymore in Firefox (tested with Firefox v102 and v99) and pipewire 0.3.52. With wf-recorder v0.2.1 it worked.

I can't reproduce the issue on Chromium v103 on another machine though, so I guess it's a Firefox issue? I've diffed the code changes for `0.2.1..0.3.0` but didn't immediately identify the reason. 

By diffing the output of `ffmpeg -i` on two test files I see this (not sure if relevant):
```
# (v0.2.1) wf-recorder -f out.mp4
Stream #0:0(und): Video: h264 (Constrained Baseline) (avc1 / 0x31637661), yuv420p, 786x220, 122 kb/s, 60.09 fps, 60 tbr, 90k tbn, 120 tbc (default)

# (0.3.0) wf-recorder -f out.mp4
Stream #0:0(und): Video: h264 (High 4:4:4 Predictive) (avc1 / 0x31637661), yuvj444p(pc), 2560x1440 [SAR 1:1 DAR 16:9], 6135 kb/s, 6.05 fps, 90k tbr, 90k tbn, 2000k tbc (default)
```

I'm fine with handtuning `wf-recorder` to restore the previous behaviour if I knew which knob to use. Any suggestion? 

thanks

soreau:
Does `wf-recorder -t -f out.mp4` help?

apiraino:
fantastic, It works! :)

Now, just out of curiosity, can you also detail why? Why is the profile `High 4:4:4 Predictive` problematic or is that an issue of Firefox? What's the actual effect of "[...] convert the data in software, before sending it to the GPU
" (as pr the man page)?

By inspecting the two files metadata with `mediainfo` I only see this perhaps relevant difference:
```
# Baseline
< Color space                              : YUV
< Chroma subsampling                       : 4:2:0

# Hi444pp
> Chroma subsampling                       : 4:4:4
```

And by looking at the [h264 profiles wikipedia page](https://en.wikipedia.org/wiki/H.264/MPEG-4_AVC#Profiles) I can infer that the "Constrained Baseline" is a kind of a common denominator, the "High 4:4:4 Predictive" must have something that not all browsers support. Makes sense?

soreau:
> fantastic, It works! :)

Glad to hear it.

> 
[...]

PR #237, opened 2023-09-18
Set the default pixel format to yuv420p

### LINKED ISSUE #152: All recorded videos appear to be corrupt (?)
I record videos with this script:

```bash
#!/bin/bash

# Convenience script for recording under sway

if [[ -z "$1" ]]; then
    echo "Fatal error: First argument must be the recording destionation" >&2
    exit 1
fi

wf-recorder \
    -g "$(slurp-window)" \
    -c libx264rgb \
    -p "tune=animation" \
    -p "preset=faster" \
    -p "crf=14" \
    --audio="alsa_output.pci-0000_11_00.4.analog-stereo.monitor" \
    --file "$1"
```

The resulting video plays well with `mpv` and VLC, but if I try to embed it in a browser it doesn't work:

https://user-images.githubusercontent.com/597206/119030219-e3f49280-b9a9-11eb-8ca7-cce5924f6111.mp4

If I try to open the file in Avidemux to re-encode it, all it shows is a green box:

image

Does anybody else have this issue? Is it an issue with wf-recorder or with my ffmpeg parameters?

soreau:
Only when using the default codec and the pixel format is not set.
This makes the videos more web friendly by default.

Closes #152.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -1,6 +1,7 @@
 #pragma once
 
 #define DEFAULT_CODEC "@default_codec@"
+#define DEFAULT_PIX_FMT "@default_pix_fmt@"
 #define DEFAULT_AUDIO_CODEC "@default_audio_codec@"
 #define DEFAULT_AUDIO_SAMPLE_RATE @default_audio_sample_rate@
 #define DEFAULT_CONTAINER_FORMAT "@default_container_format@"
--- a/meson.build
+++ b/meson.build
@@ -16,6 +16,7 @@ project(
 conf_data = configuration_data()
 
 conf_data.set('default_codec', get_option('default_codec'))
+conf_data.set('default_pix_fmt', get_option('default_pixel_format'))
 conf_data.set('default_audio_codec', get_option('default_audio_codec'))
 conf_data.set('default_audio_sample_rate', get_option('default_audio_sample_rate'))
 conf_data.set('default_container_format', get_option('default_container_format'))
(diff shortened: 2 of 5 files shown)


nowrep:
That's not a good idea because `params.pix_fmt` is used as override. This will break vaapi encoding.

https://github.com/ammen99/wf-recorder/blob/22f18cdf481bb98bced3fc40baa52f877ea4272f/src/frame-writer.cpp#L181-L186

soreau:
> That's not a good idea because `params.pix_fmt` is used as override. This will break vaapi encoding and possibly other encoders too.
> 
> https://github.com/ammen99/wf-recorder/blob/22f18cdf481bb98bced3fc40baa52f877ea4272f/src/frame-writer.cpp#L181-L186

What's wrong with setting `-x ""`? You already have to set additional options to use vaapi, and this could be documented in the wiki or so.

nowrep:
You only need to set codec eg. `-c h264_vaapi`.

> What's wrong with setting -x ""?
 
You can also set `-x yuv420p` if you don't want higher quality yuv444p which would be the default. Another issue being if you use `render_bit_depth 10` it automatically picks 10bit format yuv444p10 now, but with this change it would use 8bit format.

soreau:
> You only need to set codec eg. `-c h264_vaapi`.
> 
> > What's wrong with setting -x ""?
> 
> You can also set `-x yuv420p` if you don't want higher quality yuv444p which would be the default. Another issue being if you use `render_bit_depth 10` it automatically picks 10bit format yuv444p10 now, but with this change it would use 8bit format.

The patch allows you to set the default when configuring with meson. You can set option `-Ddefault_pixel_format=""` if you want the old behavior, but the default should work for users to submit to the web easily. Just look at issues #218, #189, #175 and #152.

ammen99:
Can we use this option only if we use libx264? We have preset options for particular codecs, I don't see why we shouldn't set yuv420p as default in those cases, and if the user goes on to do something more special, we wont get in their way.

soreau:
@ammen99 @nowrep I've updated the patch, are there still concerns?

ammen99 (approved):
Looks good now, thanks!

PR #242, opened 2023-11-18
Get hw_frame_context from buffersink

nowrep:
Fixes VAAPI recording in formats other than NV12 (eg. P010).

```
wf-recorder -F scale_vaapi=format=p010:out_range=full -c hevc_vaapi
```
video format is now correctly set to `yuv420p10` instead of `yuv420p`.


ammen99 (approved):
Seems to work fine here, and I suppose it makes sense .. Thanks :)

PR #247, opened 2023-12-12
PipeWire audio capture

nowrep:


src/pipewire.cpp
```diff
@@ -0,0 +1,187 @@
...
+            } else {
+                pw_properties_set(props, PW_KEY_STREAM_CAPTURE_SINK, source_is_sink ? "true" : "false");
+                pw_properties_set(props, PW_KEY_TARGET_OBJECT, params.audio_source);
+            }
+            spa_hook_remove(&registry_listener);
+            pw_proxy_destroy(reinterpret_cast<struct pw_proxy*>(registry));
+        }
+    }
+
+    stream = pw_stream_new(core, "wf-recorder3", props);
```
soreau, line 166: Is "wf-recorder3" a typo or intentional for some reason?

src/audio.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "pipewire.hpp"
+#endif
+
+AudioReader *AudioReader::create(AudioReaderParams params)
+{
+#ifdef HAVE_PIPEWIRE
+    AudioReader *pw = new PipeWireReader;
+    pw->params = params;
+    if (pw->init())
+        return pw;
```
soreau, line 18: This seems to silently choose pipewire by default and use pulse as a fallback. What if both are built-in and both available, but the user wants to select pulse?

soreau (commented):
The idea looks good to me, I just have a couple questions.

src/pipewire.cpp
nowrep: Copied it to match the pulse code, but not sure if it actually isn't a typo there.

https://github.com/ammen99/wf-recorder/blob/c43e8e11ec99e72d8c137fa2ea7e4ddb934275bd/src/pulse.cpp#L29-L30

src/audio.cpp
nowrep: If user is running pipewire I'd say there is no reason to choose the pulse backend.

soreau: It could be merely for testing, if pipewire backend isn't working or if the user has a preference.

soreau: maybe it could be an env var instead of an option to make it easy?

src/pipewire.cpp
soreau: It probably could stand to be corrected in both places.

src/audio.cpp
nowrep: On my system the pulse backend is completely broken when using the default output selection (just `-a`), it hangs wf-recorder and while the process is running all i hear is buzzing sound, and also minor annoyance is that for capturing sinks it requires monitors. None of these issues are with pipewire backend, default audio selection will also automatically switch when changing default device during recording and it also supports capturing single application. So I can't really imagine anyone wanting that instead of proper working audio capture.

soreau: Are you saying that wf-recorder current git without any patching has completely broken pulseaudio support? This would be quite surprising to me since 1) users have not reported such problems with pulseaudio recording and 2) I run pipewire with default wf-recorder git and there is no such issues. This makes me wonder if your patches break pulseaudio support in the mix.

nowrep: Yeah, of course I mean unpatched, can also reproduce on wf-recorder from Arch repo. May just be my system then.

> This makes me wonder if your patches break pulseaudio support in the mix.

You can try to build from this branch :)

soreau: Yes, probably your system. But it would be nice to have a runtime option to select pulseaudio instead of pipewire if both are built in and available. It's not as easy if you have to rebuild wf-recorder with `-Dpipewire=disabled`, just to force pulseaudio. Lastly, I don't think it's fair to say that users don't have a preference in this context. The other thing I notice is that it is more likely that pipewire causes wf-recorder to fail to build. First thing I notice is that there is no version requirement in the meson.build for pipewire dependency. Here, it fails to build here with `/usr/include/pipewire-0.3/pipewire/properties.h:15:10: fatal error: spa/utils/string.h: No such file or directory` (despite /usr/include/spa-0.2/spa/utils/string.h presence) and while this might be some pipewire fail, I think pipewire should be disabled by default.

nowrep: Can you post full build log? `libspa-0.2` is dependency of `libpipewire-0.3` so it should get picked up automatically.

src/pipewire.cpp
nowrep: I'd keep it to not break things.

src/audio.cpp
apiraino: > If user is running pipewire I'd say there is no reason to choose the pulse backend.

my .2 cents as a user is that I would be surprised if wf-recorder changed some important default without bumping major version (at least).

Also, I'd be happy to use the pipewire backend and remove the PulseAudio replacement daemon, if pipewire worked flawlessy OOTB (which is not my case unfortunately)

nowrep: > if pipewire worked flawlessy OOTB

Does this PR work for you or not? Otherwise I don't see how that's relevant here.

apiraino: (sorry for jumping into this thread of comments)

what I meant is that I'd love to switch to the Pipewire backend without looking back but currently that's for me not possible, so I assume also other users like me run both Pulse and Pipewire at the same time. Having `wf-recorder` switch to the Pipewire backend could be a breaking change. Happy to helpt testing this patch before merge, though!

In short: please avoid breaking changes if possible

ammen99 (commented):
Overall LGTM, would you consider adding something like I have in my patch here: https://0x0.st/Xba8.diff ? 

It makes the audio/video sync a bit better in my limited tests.

nowrep:
> would you consider adding something like I have in my patch here: https://0x0.st/Xba8.diff ?

Not sure what the equivalent for pipewire would be, but there is a `pw_buffer.time` with a timestamp for every buffer which I think should be better than static offset?

ammen99:
> > would you consider adding something like I have in my patch here: https://0x0.st/Xba8.diff ?
> 
> Not sure what the equivalent for pipewire would be, but there is a `pw_buffer.time` with a timestamp for every buffer which I think should be better than static offset?

As far as I could see in frame-writer.cpp we do not use audio timestamps - so it all boils down to figuring out what is the offset of a particular video frame relative to the start of the audio stream (which I assume consists of packed frames without any 'holes' between them?)

nowrep:
> so it all boils down to figuring out what is the offset of a particular video frame relative to the start of the audio stream

In that case maybe wait for first audio frame while dropping early video frames? 

<details>
  <summary>0001-pw-sync.patch</summary>
  
  ```patch
From 6cc0f969504b592532665934117527c72205acd6 Mon Sep 17 00:00:00 2001
From: David Rosca <nowrep@gmail.com>
Date: Wed, 19 Jun 2024 10:32:59 +0200
Subject: [PATCH] pw sync

---
 meson.build      |  2 +-
 src/audio.hpp    |  1 +
 src/main.cpp     | 65 +++++++++++++++++++++++++++++-------------------
 src/pipewire.cpp |  8 ++++++
 src/pipewire.hpp |  3 +++
 5 files changed, 53 insertions(+), 26 deletions(-)

diff --git a/meson.build b/meson.build
index 415271f..46e16b1 100644
--- a/meson.build
+++ b/meson.build
@@ -6,7 +6,7 @@ project(
 	license: 'MIT',
 	meson_version: '>=0.54.0',
 	default_options: [
-		'cpp_std=c++11',
+		'cpp_std=c++17',
         'c_std=c11',
 		'warning_level=2',
 		'werror=false',
diff --git a/src/audio.hpp b/src/audio.hpp
index fd817b1..ce5425c 100644
--- a/src/audio.hpp
+++ b/src/audio.hpp
@@ -20,6 +20,7 @@ public:
     virtual void start() = 0;
     AudioReaderParams params;
     static AudioReader *create(AudioReaderParams params);
+    virtual uint64_t get_time_base() const { return 0; }
 };
 
 #endif /* end of include guard: AUDIO_HPP */
diff --git a/src/main.cpp b/src/main.cpp
index 0293041..92cd812 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,6 +1,7 @@
 #define _XOPEN_SOURCE 700
 #define _POSIX_C_SOURCE 199309L
 #include <iostream>
+#include <optional>
 
 #include <list>
 #include <string>
@@ -564,6 +565,8 @@ static void write_loop(FrameWriterParams params)
     std::unique_ptr<AudioReader> pr;
 #endif
 
+    std::optional<uint64_t> first_frame_ts;
+
     while(!exit_main_loop)
     {
         // wait for frame to become available
@@ -604,28 +607,48 @@ static void write_loop(FrameWriterParams params)
 #endif
         }
 
+        bool drop = false;
+        uint64_t sync_timestamp = 0;
+        if (first_frame_ts.has_value()) {
+            sync_timestamp = buffer.base_usec - first_frame_ts.value();
+        } else if (pr) {
+            if (!pr->get_time_base() || pr->get_time_base() > buffer.base_usec) {
+                drop = true;
+            } else {
+                first_frame_ts = pr->get_time_base();
+                sync_timestamp = buffer.base_usec - first_frame_ts.value();
+            }
+        } else {
+            sync_timestamp = 0;
+            first_frame_ts = buffer.base_usec;
+        }
+
         bool do_cont = false;
 
-        if (use_dmabuf) {
-            if (use_hwupload) {
-                uint32_t stride = 0;
-                void *map_data = NULL;
-                void *data = gbm_bo_map(buffer.bo, 0, 0, buffer.width, buffer.height,
-                    GBM_BO_TRANSFER_READ, &stride, &map_data);
-                if (!data) {
-                    std::cerr << "Failed to map bo" << std::endl;
-                    break;
+        if (!drop) {
+            if (use_dmabuf) {
+                if (use_hwupload) {
+                    uint32_t stride = 0;
+                    void *map_data = NULL;
+                    void *data = gbm_bo_map(buffer.bo, 0, 0, buffer.width, buffer.height,
+                        GBM_BO_TRANSFER_READ, &stride, &map_data);
+                    if (!data) {
+                        std::cerr << "Failed to map bo" << std::endl;
+                        break;
+                    }
+                    do_cont = frame_writer->add_frame((unsigned char*)data,
+                        sync_timestamp, buffer.y_invert);
+                    gbm_bo_unmap(buffer.bo, map_data);
+                } else {
+                    do_cont = frame_writer->add_frame(buffer.bo,
+                        sync_timestamp, buffer.y_invert);
                 }
-                do_cont = frame_writer->add_frame((unsigned char*)data,
-                    buffer.base_usec, buffer.y_invert);
-                gbm_bo_unmap(buffer.bo, map_data);
             } else {
-                do_cont = frame_writer->add_frame(buffer.bo,
-                    buffer.base_usec, buffer.y_invert);
+                do_cont = frame_writer->add_frame((unsigned char*)buffer.data,
+                    sync_timestamp, buffer.y_invert);
             }
         } else {
-            do_cont = frame_writer->add_frame((unsigned char*)buffer.data,
-                buffer.base_usec, buffer.y_invert);
+            do_cont = true;
         }
 
         frame_writer_mutex.unlock();
@@ -1239,9 +1262,6 @@ int main(int argc, char *argv[])
 
     printf("selected region %d,%d %dx%d\n", selected_region.x, selected_region.y, selected_region.width, selected_region.height);
 
-    timespec first_frame;
-    first_frame.tv_sec = -1;
-
     bool spawned_thread = false;
     std::thread writer_thread;
 
@@ -1280,12 +1300,7 @@ int main(int argc, char *argv[])
             spawned_thread = true;
         }
 
-        if (first_frame.tv_sec == -1)
-            first_frame = buffer.presented;
-
-        buffer.base_usec = timespec_to_usec(buffer.presented)
-            - timespec_to_usec(first_frame);
-
+        buffer.base_usec = timespec_to_usec(buffer.presented);
         buffers.next_capture();
     }
 
diff --git a/src/pipewire.cpp b/src/pipewire.cpp
index 0835861..9f213d1 100644
--- a/src/pipewire.cpp
+++ b/src/pipewire.cpp
@@ -110,6 +110,9 @@ static void on_stream_process(void *data)
         }
     }
 
+    if (!pr->time_base)
+        pr->time_base = b->time;
+
     pw_stream_queue_buffer(pr->stream, b);
 }
 
@@ -203,3 +206,8 @@ void PipeWireReader::start()
 
     pw_thread_loop_unlock(thread_loop);
 }
+
+uint64_t PipeWireReader::get_time_base() const
+{
+    return time_base / 1000;
+}
diff --git a/src/pipewire.hpp b/src/pipewire.hpp
index a9943c9..dccbcd3 100644
--- a/src/pipewire.hpp
+++ b/src/pipewire.hpp
@@ -11,6 +11,7 @@ public:
     ~PipeWireReader();
     bool init() override;
     void start() override;
+    uint64_t get_time_base() const override;
 
     struct pw_thread_loop *thread_loop = nullptr;
     struct pw_context *context = nullptr;
@@ -25,6 +26,8 @@ public:
 
     uint8_t *buf = nullptr;
     size_t buf_size = 0;
+
+    uint64_t time_base = 0;
 };
 
 #endif /* end of include guard: PIPEWIRE_HPP */
-- 
2.45.2


  ```
</details>

ammen99:
@nowrep Looks good

ammen99:
@nowrep Will you push your patch to this branch? Then I will test & merge

ammen99:
Also please make the pulse backend default for now, apparently it doesn't work for everyone and I expect a long time to pass until pipewire works reliably across all distros.

nowrep:
Pushed. For pulse it will need the change from your patch (implement `get_time_base`).

src/audio.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "pulse.hpp"
+#endif
+
+#ifdef HAVE_PIPEWIRE
+#include "pipewire.hpp"
+#endif
+
+AudioReader *AudioReader::create(AudioReaderParams params)
+{
+#ifdef HAVE_PIPEWIRE
```
ammen99, line 14: This got me thinking, if we eventually want/need to support even more audio backends (like ALSA), maybe we should add a new command line argument for selecting the audio capture backend? It could be a long option only.

ammen99:
@nowrep I pushed two commits, one is an implementation of get_time_base() for pulseaudio and another one which makes it possible to select the default backend at runtime, as well as at configure time. Please take a look, I hope I didn't break something ;)

nowrep:
Looks good to me :+1:

ammen99 (approved):
Great, thanks!

0xVasconcelos:
I tried to compile from master with libpipewire-0.3-dev installed in Ubuntu 23.10 and got this

```
2.384 [15/20] Compiling C++ object wf-recorder.p/src_audio.cpp.o
2.576 [16/20] Compiling C++ object wf-recorder.p/src_pipewire.cpp.o
2.576 FAILED: wf-recorder.p/src_pipewire.cpp.o
2.576 c++ -Iwf-recorder.p -I. -I.. -I/usr/include/x86_64-linux-gnu -I/usr/include/libdrm -I/usr/include/pipewire-0.3 -I/usr/include/spa-0.2 -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -std=c++17 -O3 '-DWFRECORDER_VERSION="0.4.1-9978c35 (" __DATE__ ", branch '"'"'master'"'"')"' -Wno-deprecated-declarations -D_REENTRANT -pthread -MD -MQ wf-recorder.p/src_pipewire.cpp.o -MF wf-recorder.p/src_pipewire.cpp.o.d -o wf-recorder.p/src_pipewire.cpp.o -c ../src/pipewire.cpp
2.576 ../src/pipewire.cpp: In function 'void on_stream_process(void*)':
2.576 ../src/pipewire.cpp:114:28: error: 'struct pw_buffer' has no member named 'time'
2.576   114 |         pr->time_base = b->time;
2.576       |                            ^~~~
2.621 [17/20] Compiling C++ object wf-recorder.p/src_pulse.cpp.o
3.159 [18/20] Compiling C++ object wf-recorder.p/src_main.cpp.o
3.166 [19/20] Compiling C++ object wf-recorder.p/src_frame-writer.cpp.o
3.166 ninja: build stopped: subcommand failed.```

0xVasconcelos:
Forget it, the lib was too old in 23.10 repos, updated to oracular and now works! :p

PR #249, opened 2024-02-09
Add -q,--quit-after <seconds> option to limit recording time

warrenfalk:
Awesome project

This pull request would add a "--quit-after <seconds>" option that allows you to simply limit the recording time automatically.

Among other things, this allows you to trivially create scripts which always recording to a rotating filename, thus allowing a cleanup process to remove old ones kind of like a tape loop.

### DIFF
--- a/manpage/wf-recorder.1
+++ b/manpage/wf-recorder.1
@@ -16,6 +16,7 @@
 .Op Fl -no-dmabuf
 .Op Fl D, -no-damage
 .Op Fl f Ar filename.ext
+.Op Fl q, -quit-after Ar seconds
 .Op Fl F Ar filter_string
 .Op Fl g, -geometry Ar geometry
 .Op Fl h, -help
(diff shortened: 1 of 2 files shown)


soreau:
It is trivial to send SIGINT via script, so this option isn't strictly necessary.

warrenfalk:
Can you tell me how to do it from bash, for example, so that you don't wait the full duration if it exits early and can furthermore react to nonzero exit codes?  I suppose trivial is subjective, but I made this change when it seemed to me that this is a difficult problem to solve for most people.

ammen99:
Also I guess what @soreau proposes is a script like this;

```sh
#!/bin/sh
wf-recorder <whatever params> &
pid = $!
sleep X
kill -INT $pid
```

warrenfalk:
Right, but the problem with that script is if libx264 crashes for some reason (which it does), that sleep keeps sleeping

ammen99:
> Right, but the problem with that script is if libx264 crashes for some reason (which it does), that sleep keeps sleeping

You can use timeout(1)

ammen99:
So the script would be like this:

```sh
timeout --signal=INT <X seconds> -- wf-recorder <whatever params>
```

I originally completely forgot that there is a timeout which does exactly what you need here xD

warrenfalk:
Ah, `timeout`. OK

ammen99:
Thanks for the PR nonetheless :)

PR #255, opened 2024-03-01
fix invalid conditional statements at line 240.

MapoMagpie:
recreate-on-format-change bf17503987995b4335ffc7764a5d04e0502e8328


ammen99 (approved):
thanks ;)

PR #256, opened 2024-03-05
exit on bad geometry

dacyberduck:
When geometry argument received is invalid, prevent wf-recorder from recording rather than setting geometry to default '0,0 0x0'

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -728,15 +728,15 @@ struct capture_region
     capture_region(int32_t _x, int32_t _y, int32_t _width, int32_t _height)
         : x(_x), y(_y), width(_width), height(_height) { }
 
-    void set_from_string(std::string geometry_string)
+    bool set_from_string(std::string geometry_string)
     {
         if (sscanf(geometry_string.c_str(), "%d,%d %dx%d", &x, &y, &width, &height) != 4)
         {
-            fprintf(stderr, "Bad geometry: %s, capturing whole output instead.\n",
+            fprintf(stderr, "Bad geometry: %s\n",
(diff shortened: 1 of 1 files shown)


soreau:
Maybe have the output message suggest that the geometry might need to be in quotation marks, too?

dacyberduck:
> Maybe have the output message suggest that the geometry might need to be in quotation marks, too?

That is already mentioned in the documentation as well as '--help' text. Adding something like 'see: wf-recorder --help' to the message should suffice.

soreau:
Maybe `wf-recorder --help | grep geometry` or even more direct: `Bad value of geometry: %s\nDid you forget to wrap it in quotation marks?`

dacyberduck:
> Maybe `wf-recorder --help | grep geometry` or even more direct: `Bad value of geometry: %s\nDid you forget to wrap it in quotation marks?`

Well, we can't simply assume that the value is wrong only because it is not properly quoted. Maybe the user has given a completely wrong data that doesn't represent any geometry.

soreau:
The reason I suggest the message suggestion is because 1) the number one reason this happens is because people forget the quotations and 2) a message telling people to look at `--help` output isn't very useful when it's not clearly spelled out that quotations are needed there. Maybe we could combine the ideas and suggest both `--help` and ask if the argument is not quoted?

ammen99 (changes requested):
I agree with @soreau, it would be nice to directly print an example of how to set the geometry and say it should be quoted.

taniusha1232 (changes requested):
/start

PR #262, opened 2024-05-06
Add compatibility with FFMPEG 7.0

eclipseo:
channel_layout has been replaced with ch_layout


ainola:
@ammen99 We're going to ship this patch in Arch as we're moving to ffmpeg 7 shortly.

ammen99 (approved):
Thanks!

Issue #174, opened 2022-06-21
Overwrite existing recording without asking


apiraino:
Hi, I've upgraded to wf-recorder 0.3.0 and now I can't overwrite anymore an existing recording.

I understand the use case of #115 but I have the opposite need :) In addition since I use Sway and often have two monitors attached I [script wf-recorder as follows](https://github.com/ammen99/wf-recorder/issues/124#issuecomment-1110321983):
```
# On multi-monitor setup, retrieve the one with focus (ex. "DP-1")
output_name=$(swaymsg -t get_outputs | jq -r '.[] | select(.focused ).name')
output_id=$(
    false | wf-recorder 2>&1 | grep $output_name | cut -d"." -f 1
)
echo -n $output_id | wf-recorder -f ~/tmp/out.mp4
```

Now I can't do this anymore. Is there a way to also add a `yes` somewhere to skip the overwrite prompt? Or an additional patch to add a `--force` parameter would be needed?

As a workaround I can record to ``~/tmp/`date +%Y%m%d-%H%M%S`-out.mp4`` ...

Thanks for wf-recorder!

ammen99:
Try something like `printf "Y\n$output_id\n" | wf-recorder -f ~/tmp/out.mp4`

ammen99:
I guess you'd also want to not print Y if the file would not be overwritten, huh. It might be a good idea to add an option for this

apiraino:
I think I can live with removing the file prior to running wf-recording but if you prefer I can submit a patch to add a an additional parameter (`-O --overwrite-output`) for evaluation (no pressure to merge)

soreau:
Does `wf-recorder -f ~/tmp/out.mp4 <<< Y` work?

apiraino:
> Does `wf-recorder -f ~/tmp/out.mp4 <<< Y` work?

that was a nice trick, but `echo -n $output_id | wf-recorder -f ~/tmp/out.mp4 <<<Y` seems it's not working

soreau:
How about `printf '%s\n' Y $output_id | wf-recorder -f ~/tmp/out.mp4`?

soreau:
Actually, `wf-recorder -o $output_name -f ~/tmp/out.mp4 <<< Y`, which eliminates the need to translate output to numeric representation, if I understand correctly what you are doing.

apiraino:
@soreau Can't believe I've overlooked that param for so long :facepalm: 

The correct solutions are:
```
[...]

PR #266, opened 2024-06-23
add -O/--overwrite

nekopsykose:
i prefer this over echoing stuff to stdin like suggested in https://github.com/ammen99/wf-recorder/issues/174

up to you though, i don't mind if it's rejected :)


ammen99 (commented):
Many package managers provide `-y`, maybe that would be better?

nekopsykose:
that sounds good to me too. no preference here

ammen99:
I'd say then make it a `-y`, let's keep to the conventions ;)

nekopsykose:
at the time i wrote that in the middle of the night i forgot what letters of the alphabet were still available... :D

ammen99 (approved):
Thanks!

PR #277, opened 2024-09-11
fix compilation when building w/o audio support

sevz17:


src/main.cpp
```diff
@@ -1136,7 +1140,11 @@ int main(int argc, char *argv[])
                 break;
 
             case '*':
+#ifdef HAVE_AUDIO
                 audioParams.audio_backend = optarg;
+#else
+                std::cerr << "Ignoring --audio-backend. Built without audio support." << std::endl;
```
sevz17, line 1146: Alternatively, it could be a hard error.

ammen99: Hey, sorry for the late reply. Yeah I think a hard error would be great! Otherwise LGTM.

ammen99 (commented):
Hey, I just merged another PR doing the same thing ;( I had forgotten about this PR. Sorry about that. In any case, thanks for your contribution!

PR #280, opened 2024-09-16
frame-writer: Use superfast instead of ultrafast for libx264/5

soreau:
This aims to default videos to use High profile by default instead of Constrained Baseline. This reportedly has more web compatibilty and makes video files slightly smaller.


ammen99 (approved):
Thanks for investigating this!

Issue #127, opened 2020-12-08
"Bad geometry" for use with slurp


nwwt:
Trying `wf-recorder -g $(slurp)` only leads to `Bad geometry: 0,0, capturing whole output instead.` with whatever region selected (the two numbers varying accordingly ofc), including the entire screen (a single QHD one).

*edit:* slurp on its own seems to return its ordinary format: `0,0 2560x1440`, `520,482 948x451`, etc.

*edit2:* Uhm, awkward. C&Ping slurp's output directly instead of via variable works as intended‽

[Arch, sway 1.5.1 / wlroots 0.12.0]

nwwt:
Stupid me, I missed quoting the variable. Sorry for the noise.

magdesign:
i have the feeling that when trying to set the geometry with a variable, which is executed prior with slurp, wf-recorder will not catch the geometry, even with added double quotes....

o-alquimista:
> Stupid me, I missed quoting the variable. Sorry for the noise.

It's not your fault. The man page `wf-recorder(1)` has it unquoted in its examples:

> EXAMPLES
>
> To select a specific part of the screen you can either use --g geometry or use https://github.com/emersion/slurp for interactive selection of the screen area that will be recorded:
>
> $ wf-recorder -g $(slurp)

soreau:
PR's welcome. 👍

o-alquimista:
Created PR #290

PR #290, opened 2024-12-15
Update wf-recorder.1: fix quote rendering in slurp call

o-alquimista:
Addresses issue #127. See this [comment](https://github.com/ammen99/wf-recorder/issues/127#issuecomment-2543977778).


ammen99 (approved):
Thanks!

PR #295, opened 2025-02-08
Add fish shell completions

floflo0:


ammen99 (changes requested):
LGTM, just please put the completion file in a new directory called `completions`, just so that we can put other completion files there in the future.

Issue #287, opened 2024-11-22
[BUG] WebM with Opus audio no sound


sakarie9:
### WebM with opus

`wf-recorder -f v.webm -c libvpx -a -C libopus`

<details>
<summary>wf-recorder output</summary>

```
Setting codec option: cpu-used=5
Setting codec option: deadline=realtime
Using video filter: null
[libvpx @ 0x7d773c001c80] v1.14.1
[libvpx @ 0x7d773c001c80] Neither bitrate nor constrained quality specified, using default CRF of 32 and bitrate of 256kbit/sec
Setting codec option: cpu-used=5
Setting codec option: deadline=realtime
Choosing sample format s16 for audio codec libopus
[libopus @ 0x7d773c4ff1c0] No bit rate set. Defaulting to 96000 bps.
Output #0, webm, to 'v.webm':
  Stream #0:0: Video: vp8, yuv420p(pc), 1920x1080 [SAR 1:1 DAR 16:9], q=2-31, 256 kb/s
      Side data:
        cpb: bitrate max/min/avg: 0/0/0 buffer size: 0 vbv_delay: N/A
  Stream #0:1: Audio: opus, 48000 Hz, stereo, s16, 96 kb/s
Using PulseAudio device: default
[Source @ 0x7d773c004fc0] Changing video frame properties on the fly is not supported by all filters.
[Source @ 0x7d773c004fc0] filter context - w: 1920 h: 1080 fmt: 119 csp: gbr range: unknown, incoming frame - w: 1920 h: 1080 fmt: 119 csp: unknown range: unknown pts_time: 0.008543
```

</details>

<details>
<summary>mediainfo output</summary>

```
General
Complete name                            : v.webm
Format                                   : WebM
Format version                           : Version 4
File size                                : 251 KiB
Duration                                 : 3 s 468 ms
Overall bit rate                         : 592 kb/s
Writing application                      : Lavf61.7.100
Writing library                          : Lavf61.7.100

Video
ID                                       : 1
Format                                   : VP8
Codec ID                                 : V_VP8
Duration                                 : 3 s 234 ms
Width                                    : 1 920 pixels
Height                                   : 1 080 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Variable
Compression mode                         : Lossy
Default                                  : No
Forced                                   : No
Color range                              : Full

Audio
ID                                       : 2
Format                                   : Opus
Codec ID                                 : A_OPUS
[...]

PR #298, opened 2025-03-03
Fix silent audio when recoding webm + opus

ThomasBaruzier:
This MR fixes https://github.com/ammen99/wf-recorder/issues/287

There is no `[libopus @ 0x7a681021cf40] Queue input is backward in time` warning when recording.

I did not test extensively this fix regarding other codecs and audio formats.


soreau (approved):
This seems like a good solution according to everything I've read. It fixes the problem without the extra libopus warning messages and printing the audio pts values increment by larger values, 960 instead of 20 in the case of vpx+libopus and 1024 in the default mp4 case. 👍

ammen99 (approved):
thanks :)

PR #302, opened 2025-04-16
Include `--file` on the manpage

Managor:
`--file` is documented at the end of the man page but isn't listed along with the other options.


ammen99 (approved):
thanks ;)

Issue #259, opened 2024-04-19
Documentation unclear on how -p option works.


Rabcor:
```
  -p, --codec-param         Change the codec parameters.
                            -p <option_name>=<option_value>
```

Is not descriptive enough

I have these settings for my codec
`-tier high -spatial-aq 1 -multipass fullres -tune:v hq -qmin 20 -rc:v vbr_minqp -cq:v 29 -preset:v p7`
and I have no idea how to translate it to` -p <option_name>=<option_value>`

i mean is it supposed to be:
`-p "-tier high -spatial-aq 1 -multipass fullres -tune:v hq -qmin 20 -rc:v vbr_minqp -cq:v 29 -preset:v p7"`

Or is it supposed to be:
`-p "tier=high spatial-aq=1 multipass=fullres tune:v=hq qmin=20 rc:v=vbr_minqp cq:v=29 preset:v=p7"`

Or is it even:
`-p tier=high -p spatial-aq=1 -p multipass=fullres -p tune:v=hq -p qmin=20 -p rc:v=vbr_minqp -p cq:v=29 -p preset:v=p7`

How does it actually work? I couldn't even find any examples of it in use.

ammen99:
> Or is it even: `-p tier=high -p spatial-aq=1 -p multipass=fullres -p tune:v=hq -p qmin=20 -p rc:v=vbr_minqp -p cq:v=29 -p preset:v=p7`

This is the correct format ;) Feel free to send a PR with an example.

Rabcor:
> > Or is it even: `-p tier=high -p spatial-aq=1 -p multipass=fullres -p tune:v=hq -p qmin=20 -p rc:v=vbr_minqp -p cq:v=29 -p preset:v=p7`
> 
> This is the correct format ;) Feel free to send a PR with an example.

Thanks.

I literally do not know how to make a PR, I mean i go to the pull request tab and I get instantly confused. github seems massively overcomplicated and I would use this so infrequently that I haven't had a real motive to learn it yet. I'm mostly here to make bug reports.

But here's my idea for a modified doc, do with it what you will.

```
  -p, --codec-param         Set a codec parameter; can be called multiple times.
                            -p <parameter1>=<value> -p <parameter2>=<value>
```

PR #306, opened 2025-05-22
Document multiple codec parameters

jakubvalenta:
And use the word "change" consistently.

Fixes: #259

Option before:

codec_param_before+40x40-width

Option after:

codec_param_after+40x40

Example before:

example_before+40x40-width

Example after:

example_after+40x40-width


ammen99 (approved):
thanks :)

Issue #307, opened 2025-05-29
[feature request] multiple audio source


vdegenne:
Here's a simple use case where I need to record my voice over a short video clip playing on the screen, therefore both the headphone's mic and the headphone output (desktop) need to be set as an input source of the recording.

Pretty confident that's not possible to do that with `wf-recorder` therefore this feature request, though it might seem a little ambitious.

If you have workarounds (besides video editing) I'm all ears.

Cheers

soreau:
I think the way to do this is outlined in [the wiki](https://github.com/ammen99/wf-recorder/wiki#recording-both-mic-input-and-application-sounds).

ammen99:
yes, what @soreau suggested is the way to solve this issue ;)

powellnorma:
@ammen99 

But what if I want to store them in separate audio tracks? This is what I have done in X11, and can be quite useful. Any workarounds for that?

ammen99:
No, there are no workarounds for this. I suppose it shouldn't be too hard to add as a feature, PRs welcome.

powellnorma:
Would it make sense to rewrite/port wf-recorder into a plugin for FFmpeg, so that it can be used as input, similar to how x11grab can be used? Would that be hard to accomplish? I think that would give a lot of freedom, as all command line options of FFmpeg could be used.

I know we can use something like `./wf-recorder -c rawvideo  -x yuv420p -m avi -f >(ffplay -f avi -) -y` (ffplay can be replaced with an arbitrary FFmpeg command), but in my tests it creates noticeable video audio delay/lag.

ammen99:
> Would it make sense to rewrite/port wf-recorder into a plugin for FFmpeg, so that it can be used as input, similar to how x11grab can be used? Would that be hard to accomplish? I think that would give a lot of freedom, as all command line options of FFmpeg could be used.
> 
> I know we can use something like `./wf-recorder -c rawvideo -x yuv420p -m avi -f >(ffplay -f avi -) -y` (ffplay can be replaced with an arbitrary FFmpeg command), but in my tests it creates noticeable video audio delay/lag.

You could totally implement an ffmpeg plugin and that would likely make wf-recorder quite obsolete. But you'd have to upstream this to ffmpeg (afaik) which is not something I am personally interested in.

PR #310, opened 2025-05-30
Implement multiple audio tracks

powellnorma:
See https://github.com/ammen99/wf-recorder/issues/307#issuecomment-2920703296 

It also adds option '-V' for verbose logging (also activated by '-l'), which shows what it (specifically the new multi-track logic) is doing.


src/debug.hpp
```diff
@@ -0,0 +1,40 @@
+#ifndef DEBUG_HPP
```
ammen99, line 1: +1 for the idea

src/frame-writer.cpp
```diff
@@ -523,94 +525,216 @@ static enum AVSampleFormat convert_codec_sample_fmt(const AVCodec *codec, std::s
...
+            exit(1);
+        }
+
+        // Create SwrContext for this stream
+        SwrContext* swrCtx = swr_alloc();
+#if HAVE_CH_LAYOUT
+        AVChannelLayout source_layout = {};
+        av_channel_layout_default(&source_layout, 2);
+
+        av_opt_set_int(swrCtx, "in_sample_rate", audioCodecCtx->sample_rate, 0);
```
ammen99, line 585: seems like most of these are duplicated in the two if-else branches, I think we should keep duplication to minimum.

src/pipewire.cpp
```diff
@@ -156,58 +142,124 @@ static const struct pw_registry_events registry_events = {
...
+
+    if (!at_least_one_connected) {
+        std::cerr << "pipewire: no audio sources connected" << std::endl;
+    }
 }
 
 uint64_t PipeWireReader::get_time_base() const
 {
-    return time_base / 1000;
+    return time_base;
```
ammen99, line 264: Maybe a stupiud question but why do we no longer need the /1000 ?

src/frame-writer.cpp
```diff
@@ -523,94 +525,216 @@ static enum AVSampleFormat convert_codec_sample_fmt(const AVCodec *codec, std::s
...
 }
 #endif
 void FrameWriter::init_codecs()
 {
     init_video_stream();
 #ifdef HAVE_AUDIO
     if (params.enable_audio)
-        init_audio_stream();
+    {
+        // Get the number of audio sources from audioParams
```
ammen99, line 733: I am not a big fan of this, is there no way to pass the information to main.cpp and from there to framewriter?

ammen99 (changes requested):
I have looked briefly through the code, a few things which stood out to me. Will be testing it later (if anyone else also reads this, more testers == better).

taniusha1232 (approved):

meson.build
```diff
@@ -42,7 +42,7 @@ include_directories(['.'])
 
 add_project_arguments(['-Wno-deprecated-declarations'], language: 'cpp')
```
taniusha1232: c

Issue #173, opened 2022-06-09
The range for partial recording of the rotation screen is incorrect


tkna91:
This phenomenon occurred to me as well.

> It also recorded the wrong part of the screen when passing the geometry via slurp. For my 90 degrees rotated screen when I wanted to record something on the bottom left, it recorded something on the top left instead.

_Originally posted by @AndreasBackx in https://github.com/ammen99/wf-recorder/issues/3#issuecomment-531589321_

Commands executed

```
$ slurp
0,2481 1080x479
$ grim -g "$(slurp)" - | swappy -f -
$ wf-recorder -g "$(slurp)" -f /tmp/capture.mp4
$ mpv /tmp/capture.mp4
```
version
```
$ pacman -Q sway wlroots wf-recorder slurp grim ffmpeg mpv
sway 1:1.7-9
wlroots 0.15.1-5
wf-recorder 0.2.1-4
slurp 1.3.2-2
grim 1.4.0-2
ffmpeg 2:5.0-7
mpv 1:0.34.1-4
$
```
https://user-images.githubusercontent.com/102382754/172766625-76c801cd-f90d-4143-94ba-a53f8576e8a0.mp4

Samq64:
My vertical monitor recordings are not rotated correctly even when recording the entire screen. I'm using the Hyprland portal but OBS works fine.

Jas-SinghFSU:
Did you find a solution for this?

tkna91:
@soreau 

> If so, can you see if [this pull request](https://github.com/ammen99/wf-recorder/pull/315) fixes it?

Sorry, I am not sure how to do that and am looking into it.

tkna91:
This problem appears to have been resolved in the latest version below.
```
$ wf-recorder --version
wf-recorder 0.5.0-664fc45 (Jul 21 2025, branch 'master')
$
```

https://github.com/user-attachments/assets/4340b788-ba46-4710-839e-5229989d2801

@soreau @ammen99  Thanks a lot. Please close it as you wish.

PR #313, opened 2025-07-13
Fix recording with rotated monitors

dotaxis:
Addresses https://github.com/ammen99/wf-recorder/issues/173

Not sure this is the best way to go about this, but open to suggestions.

Only tested with 90 and 270 degree rotations because I can't handle mirroring. I guess I could add 180 degree rotation without too much work.

Still requires `-F transpose=` to be used get the actual output file to be rotated accordingly.

`wf-recorder -F transpose=1 --transform 1 -g "107,2000 1247x532" -f ~/Videos/test.mp4` works as expected on a monitor rotated 90 degrees, and coordinates grabbed via `slurp`.

`wf-recorder -F transpose=2 --transform 3 -g "107,2000 1247x532" -f ~/Videos/test.mp4` works the same when using 270 degree rotation.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -833,7 +833,7 @@ struct capture_region
 {
     int32_t x, y;
     int32_t width, height;
-    int16_t rotate_coordinates = 0;
+    int transform = WL_OUTPUT_TRANSFORM_NORMAL;
 
     capture_region()
         : capture_region(0, 0, 0, 0) {}
@@ -842,37 +842,24 @@ struct capture_region
         : x(_x), y(_y), width(_width), height(_height) { }
 
     void set_transform(const wf_recorder_output& wo) {
-        switch (wo.transform) {
+        transform = wo.transform;
+        switch (transform) {
             case WL_OUTPUT_TRANSFORM_NORMAL:
-                fprintf(stderr, "WL_OUTPUT_TRANSFORM_NORMAL\n");
-                break;
             case WL_OUTPUT_TRANSFORM_180:
-                fprintf(stderr, "WL_OUTPUT_TRANSFORM_180\n");
-                break;
             case WL_OUTPUT_TRANSFORM_FLIPPED:
-                fprintf(stderr, "WL_OUTPUT_TRANSFORM_FLIPPED\n");
-                break;
             case WL_OUTPUT_TRANSFORM_FLIPPED_180:
-                fprintf(stderr, "WL_OUTPUT_TRANSFORM_FLIPPED_180\n");
-                break;
             case WL_OUTPUT_TRANSFORM_90:
-                rotate_coordinates = 90;
+                fprintf(stderr, "WL_OUTPUT_TRANSFORM_90\n");
                 std::swap(x, y);
                 x = -x;
-                fprintf(stderr, "WL_OUTPUT_TRANSFORM_90\n");
                 break;
             case WL_OUTPUT_TRANSFORM_270:
-                rotate_coordinates = 270;
+                fprintf(stderr, "WL_OUTPUT_TRANSFORM_270\n");
                 std::swap(x, y);
                 y = -y;
-                fprintf(stderr, "WL_OUTPUT_TRANSFORM_270\n");
                 break;
             case WL_OUTPUT_TRANSFORM_FLIPPED_90:
-                fprintf(stderr, "WL_OUTPUT_TRANSFORM_FLIPPED_90\n");
-                break;
             case WL_OUTPUT_TRANSFORM_FLIPPED_270:
-                fprintf(stderr, "WL_OUTPUT_TRANSFORM_FLIPPED_270\n");
-                break;
             default:
                 fprintf(stderr, "Transform not found.\n");
                 break;
@@ -897,26 +884,32 @@ struct capture_region
 
     bool contained_in(const capture_region& output) const
     {
-        if (rotate_coordinates == 90) {
-            return
-                output.x <= abs(x) &&
-                output.x + output.width >= x + width &&
-                output.y <= y &&
-                output.y + output.height >= y + height;
+        switch (transform) {
+            case WL_OUTPUT_TRANSFORM_180:
+            case WL_OUTPUT_TRANSFORM_FLIPPED:
+            case WL_OUTPUT_TRANSFORM_FLIPPED_180:
+            case WL_OUTPUT_TRANSFORM_90:
+                return
+                    output.x <= abs(x) &&
+                    output.x + output.width >= x + width &&
+                    output.y <= y &&
+                    output.y + output.height >= y + height;
+            case WL_OUTPUT_TRANSFORM_270:
+                return
+                    output.x <= x &&
+                    output.x + output.width >= -x + width &&
+                    output.y <= abs(y) &&
+                    output.y + output.height >= y + height;
+            case WL_OUTPUT_TRANSFORM_FLIPPED_90:
+            case WL_OUTPUT_TRANSFORM_FLIPPED_270:
+            default: //WL_OUTPUT_TRANSFORM_NORMAL
+                return
+                    output.x <= x &&
+                    output.x + output.width >= x + width &&
+                    output.y <= y &&
+                    output.y + output.height >= y + height;
         }
-        else if (rotate_coordinates == 270) {
-            return
-                output.x <= x &&
-                output.x + output.width >= -x + width &&
-                output.y <= abs(y) &&
-                output.y + output.height >= y + height;
-        }
-        return
-            output.x <= x &&
-            output.x + output.width >= x + width &&
-            output.y <= y &&
-            output.y + output.height >= y + height;
-    }
+    };
 };
 
 static wf_recorder_output* detect_output_from_region(const capture_region& region)
@@ -1173,7 +1166,7 @@ int main(int argc, char *argv[])
             case 'g':
                 selected_region.set_from_string(optarg);
                 break;
-            
+
             case 'c':
                 params.codec = optarg;
                 break;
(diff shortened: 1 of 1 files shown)


apiraino:
Just a .2 cents worth comment:

I use sway and with `swaymsg -t get_outputs` I get the current `transform` of the displays ([man page](https://man.archlinux.org/man/sway-output.5.en#:~:text=output%20%3Cname%3E%20transform%20%3Ctransform%3E%20%5Bclockwise%7Canticlockwise%5D)), which tells me if the display is rotated. Example:
```
$ swaymsg -t get_outputs 
Output eDP-1 'LG Display 0x0521 Unknown'
  Current mode: 1920x1080 @ 60.020 Hz
  Power: on
  Position: 320,1440
  Scale factor: 1.000000
  Scale filter: nearest
  Subpixel hinting: unknown
  Transform: normal
  Workspace: 1
  Max render time: off
  Adaptive sync: disabled
  Allow tearing: no
  Available modes:
    1920x1080 @ 60.020 Hz
    1920x1080 @ 47.999 Hz
```

These info are exposed by the compositor (in this case `wl_root`). I'm curious if there could be a generic way to programmatically retrieve this `transform` value of a display without having the user needing to worry about it.

dotaxis:
I actually realized right after I opened this that `grim` handles transforms automatically by polling the transform info from the compositor and that it would be much better to do that here as well.

I'll work on this using @soreau's patch above and update this when I make headway.

soreau:
[This version](https://0x0.st/8kNt.diff) of the patch removes variable names that are unused to avoid warnings.

dotaxis:
Alright, I've pushed a couple of commits to handle rotation automatically, but I will have to spin my monitor around in circles and do a bunch of testing to account for all the possible cases.

If anyone wants to build and test with 90 and 270 rotations for now, it would be much appreciated.

dotaxis:
I'm going to move the switch statement into `contained_in()` so we don't have a redundant `rotate_coordinates` value and if block, and clean some other stuff up while I'm at it. Thanks for going over it! Will ping when ready.

src/main.cpp
```diff
@@ -771,13 +833,52 @@ struct capture_region
...
     capture_region()
         : capture_region(0, 0, 0, 0) {}
 
     capture_region(int32_t _x, int32_t _y, int32_t _width, int32_t _height)
         : x(_x), y(_y), width(_width), height(_height) { }
 
+    void set_transform(const wf_recorder_output& wo) {
+        switch (wo.transform) {
+            case WL_OUTPUT_TRANSFORM_NORMAL:
+                fprintf(stderr, "WL_OUTPUT_TRANSFORM_NORMAL\n");
```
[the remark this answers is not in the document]
dotaxis: Should we still use stderr or was that just for debugging? Not sure how much info we want to give per case here. Maybe say "90 degree rotation detected, swapping x and y and using negative x" or something?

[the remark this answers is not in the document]
soreau: > Should we still use stderr or was that just for debugging?

You must use stderr. The video is output to stdout.

> Not sure how much info we want to give per case here. Maybe say "90 degree rotation detected, swapping x and y and using negative x" or something?

I was thinking something simpler along the lines of "Using transform 90 flipped" or "Using transform 90 flipped for output HDMI-A-1" if that's convenient. The idea would be to output that the transform was detected.

src/main.cpp
```diff
@@ -771,13 +833,39 @@ struct capture_region
 {
     int32_t x, y;
     int32_t width, height;
+    int transform = WL_OUTPUT_TRANSFORM_NORMAL;
```
soreau, line 836: I would move this to display_handle_geometry() or simply omit it.

src/main.cpp
```diff
@@ -771,13 +833,39 @@ struct capture_region
...
+                std::swap(x, y);
+                x = -x;
+                break;
+            case WL_OUTPUT_TRANSFORM_270:
+                fprintf(stderr, "WL_OUTPUT_TRANSFORM_270\n");
+                std::swap(x, y);
+                y = -y;
+                break;
+            case WL_OUTPUT_TRANSFORM_FLIPPED_90:
+            case WL_OUTPUT_TRANSFORM_FLIPPED_270:
```
soreau, line 862: There should be a break statement after this line, or else 90/270 flipped will fall through to the default case of 'Transform not found'.

src/main.cpp
```diff
@@ -771,13 +833,39 @@ struct capture_region
...
 
+    void set_transform(const wf_recorder_output& wo) {
+        transform = wo.transform;
+        switch (transform) {
+            case WL_OUTPUT_TRANSFORM_NORMAL:
+            case WL_OUTPUT_TRANSFORM_180:
+            case WL_OUTPUT_TRANSFORM_FLIPPED:
+            case WL_OUTPUT_TRANSFORM_FLIPPED_180:
+            case WL_OUTPUT_TRANSFORM_90:
+                fprintf(stderr, "WL_OUTPUT_TRANSFORM_90\n");
```
soreau: This message says transform 90 but it's called in cases normal, 180, 180-flipped and flipped, so the message is misleading.

src/main.cpp
```diff
@@ -470,6 +532,7 @@ static void handle_global(void*, struct wl_registry *registry,
         wf_recorder_output wro;
         wro.output = output;
         available_outputs.push_back(wro);
+        wl_output_add_listener(output, &output_listener, &available_outputs.back());
```
soreau, line 535: Maybe it would be a good idea to only add the listener on the chosen output.

soreau (changes requested):
It's shaping up, but there a few things that need attention.

src/main.cpp
dotaxis: It needs to be accessible inside the scope of `contained_in` which does not have access to `display_handle_geometry` and doesn't need to. I can omit the default value but it needs to exist there as far as I can tell.

src/main.cpp
dotaxis: These are simply not implemented yet and I plan to account for them before anything gets merged, but I can add breaks in the meantime.

src/main.cpp
```diff
@@ -771,13 +833,40 @@ struct capture_region
...
 
+    void set_transform(const wf_recorder_output& wo) {
+        transform = wo.transform;
+        switch (transform) {
+            case WL_OUTPUT_TRANSFORM_NORMAL:
+            case WL_OUTPUT_TRANSFORM_180:
+            case WL_OUTPUT_TRANSFORM_FLIPPED:
+            case WL_OUTPUT_TRANSFORM_FLIPPED_180:
+            case WL_OUTPUT_TRANSFORM_90:
+                fprintf(stderr, "90 degree rotation detected, altering geometry to compensate.\n");
```
soreau, line 852: This doesn't seem right. It says 90 but all the cases above it fall through to this message.

src/main.cpp
soreau: To show you what I meant, [here's another diff](https://0x0.st/8kmn.diff). It eliminates the need for set_transform() as well.

soreau: I don't see why [this](https://0x0.st/8kBV.diff) can't work, unless I'm still missing something.

soreau: Anyway, if you still need set_transform(), that's fine, but the transform variable for the switch would be available as `this->transform` in that patch, so no need to have it on `struct wf_recorder_output`.

soreau: I guess [this](https://0x0.st/8kMT.diff) is what you want.

> Storing transform as an 8 bit integer in capture_region is fine, isn't it?

Probably, but why not choose the enum type that was used to send it?

dotaxis: I really don't know what the goal of making all these changes is, but the result is that the transform value in capture_region is always 0. I don't know why it's happening but I'm getting tired of trying today. I'll look again tomorrow.

soreau:
@dotaxis Hi, I sort of got off into the weeds the other day but today I tested a bit and came up with [this](https://0x0.st/8kE4.diff). It pipes the transform into frame-writer.cpp, where it sets the transpose like you were doing with the -F flag. Glean what you will, hope it helps. Cheers. 👍

dotaxis:
> @dotaxis Hi, I sort of got off into the weeds the other day but today I tested a bit and came up with [this](https://github.com/soreau/wf-recorder/commit/d1a5fe41ab9d6bd399d98be605a338f05fd48173). It pipes the transform into frame-writer.cpp, where it sets the transpose like you were doing with the -F flag. Glean what you will, hope it helps. Cheers. 👍

The goal of this PR is not to apply transposition. It is to fix the handling of coordinates which do not work in a logical way when a monitor is rotated.

soreau:
> > @dotaxis Hi, I sort of got off into the weeds the other day but today I tested a bit and came up with [this](https://github.com/soreau/wf-recorder/commit/d1a5fe41ab9d6bd399d98be605a338f05fd48173). It pipes the transform into frame-writer.cpp, where it sets the transpose like you were doing with the -F flag. Glean what you will, hope it helps. Cheers. 👍
> 
> The goal of this PR is not to apply transposition. It is to fix the handling of coordinates which do not work in a logical way when a monitor is rotated.

This has already works. What version of wayfire are you using? If you can use wayfire latest git, this would be optimal.

dotaxis:
I am not using wayfire, I'm using hyprland. Nothing in this project hints at being exclusive to wayfire users.

src/main.cpp
dotaxis: Everything is not working as expected. `grim` handles coordinates from `slurp` just fine and `wf-recorder` does not.

ammen99:
@dotaxis I have tested wf-recorder with sway+slurp and coordinates work normally. Maybe Hyprland/Aquamarine do not handle the rotation correctly. In any case, since it works on Sway/Wayfire, if we include patches to make it work on Hyprland, it would break the other compositors. And this project is aimed first and foremost at wlroots-based compositors.


Unless I am mistaken, is there anyone who is having issues (beyond the video being transposed) on Sway / River / Wayfire / etc ?

soreau:
I have tested sway and wayfire with 90 rotation - both work with grim or wf-recorder using slurp. If you are having this issue, can you try with `wf-recorder --no-dmabuf` to see if it makes a difference?

PR #315, opened 2025-07-16
Automatically transpose video based on output transform

soreau:


dotaxis:
This doesn't fix #173.

The coordinates do not pass the `contained_in` check and recording fails on rotated monitors.

soreau:
I have updated the commit message and description so this will not close any issues if merged.

apiraino:
My .2 cents as an end-user: I think this is the good default I would expect. If I wanted the other way, I would probably *play* the video rotated before screen-grabbing it.

soreau:
@tkna91 I didn't want to clutter the other thread so I am responding here. I was able to reproduce this with two sway outputs, the 'first'/right-most at 90 and the second/left-most output normal. When I try wf-recorder with slurp on the 90 output, I get the same error as you. I tested and this works with master, so I will look into it and try to find out why this PR breaks it. Thanks for testing!

soreau:
I found out that the reason is because this PR switched from xdg-output to wl_output for getting the output geometry, but wlroots doesn't properly advertise the x/y coordinates, it just sends 0,0 for all outputs. This means wf-recorder is confused when using multiple outputs because it thinks they are overlapping. The wlroots devs position is that 'clients should not care' about this information and that if you want it, you must use xdg-output. This means binding both interfaces, just to get the transform for the output via wl_output and the rest via xdg-output. I will update this PR to do this when I have the gumption.

tkna91:
@soreau Rotation has normalized. However, the selection appears to be disabled.

https://github.com/user-attachments/assets/da0fb9ce-32a4-4c7a-a5c8-167c50ec4995

soreau:
@tkna91 That's great, thanks for verifying and for the videos, much appreciated! 👍

src/frame-writer.cpp
```diff
@@ -212,15 +212,62 @@ AVPixelFormat FrameWriter::handle_buffersink_pix_fmt(const AVCodec *codec)
...
+        fprintf(stderr, "Transform: 270\n");
+        return "transpose=" + std::to_string(2);
+        break;
+      case WL_OUTPUT_TRANSFORM_FLIPPED:
+        fprintf(stderr, "Transform: FLIPPED\n");
+        return "hflip";
+        break;
+      case WL_OUTPUT_TRANSFORM_FLIPPED_90:
+        fprintf(stderr, "Transform: FLIPPED_90\n");
+        return "transpose=" + std::to_string(1) + ",hflip";
```
ammen99, line 237: I thought the `FLIPPED_x` transforms mean first hflip then the transform? https://wayland.app/protocols/wayland#wl_output:enum:transform

ammen99:
Nice, thanks!

PR #317, opened 2025-07-21
feat: add output listing capability

nabiko02:
I have added the -L / --list-output flag to list available outputs for use with -o.
This allows users to programmatically get output names, useful for scripting and GUI tools like wf-recorder-gui.

### DIFF
--- a/manpage/wf-recorder.1
+++ b/manpage/wf-recorder.1
@@ -21,7 +21,7 @@
 .Op Fl h, -help
 .Op Fl l, -log
 .Op Fl m, -muxer Ar muxer
-.Op Fl L, -list-ouput
+.Op Fl L, -list-output
 .Op Fl o, -output Ar output
 .Op Fl p, -codec-param Op Ar option_param=option_value
 .Op Fl v, -version


soreau:
I don't want you to feel like you're wasting time, I'd rather see you solve your issue. I don't know how wf-recorder-gui works but maybe it could (optionally) install the required tools to use the program, such as wf-recorder and wayland-utils. Perhaps this could be a packaging dependency detail.

The other thing is that this project belongs to @ammen99, and has the final say on whether or not to include a feature such as this. So I am reopening this for @ammen99 comments.

src/main.cpp
```diff
@@ -881,6 +881,8 @@ Use Ctrl+C to stop.)");
 
   -l, --log                 Generates a log on the current terminal. Debug purposes.
 
+  -L, --list-ouput          List the available outputs.
```
ammen99, line 884:
```suggestion
  -L, --list-output          List the available outputs.
```

src/main.cpp
```diff
@@ -989,6 +991,34 @@ static void parse_codec_opts(std::map<std::string, std::string>& options, const
...
+    wl_registry_add_listener(registry, &registry_listener, NULL);
+    sync_wayland();
+}
+
+static void list_available_output()
+{
+    init_wayland_client();
+    load_output_info();
+
+    int i = 1;
```
ammen99, line 1012: I would put this into a function to really reuse it, otherwise lgtm.

ammen99 (approved):
Thanks!

PR #318, opened 2025-08-03
Fix -y/--overwrite requiring an argument

layercak3:
Otherwise it consumes the next argument which can cause wrong behaviour.


ammen99 (approved):
Thanks!

PR #324, opened 2025-10-05
Fix "Option 'pix_fmts' is not a runtime option"

### LINKED ISSUE #323: fails with cryptic error message
running `wf-recorder` in my terminal results in this output:

```
selected region 0,0 0x0
Setting codec option: crf=20
Setting codec option: preset=ultrafast
Setting codec option: tune=zerolatency
[Sink @ 0x7f081c004ac0] Option 'pix_fmts' is not a runtime option and so cannot be set after the object has been initialized
Failed to set pix_fmts: Invalid argument
```

i dont know what its trying to tell me 

it worked a couple of days ago just fine, suddenly stopped working. im on arch linux, maybe an update on a dependency broke something?

paul-ri:
Fixes #323 

This is all quite out of my depth by it looked doable thanks to knowing:
* On Arch Linux, downgrading ffmpeg to 7.1.1 was fixing the issue.
* Looking at the e-mail thread linked in the issue, someone hinted to look at ffmpeg's `doc/examples/decode_filter_video.c`. Looking at the history of the file, there were hints here and there of what could be done.

More details in the commits. Happy to hear if this wasn't an adequate fix!

### Testing done

Was able to record a video with the default parameters.

### DIFF
--- a/src/frame-writer.cpp
+++ b/src/frame-writer.cpp
@@ -335,17 +335,8 @@ void FrameWriter::init_video_filters(const AVCodec *codec)
         exit(-1);
     }
 
-    // We also need to tell the sink which pixel formats are supported.
-    // by the video encoder. codevIndicate to our sink  pixel formats
-    // are accepted by our codec.
-    const AVPixelFormat picked_pix_fmt[] =
-    {
-        handle_buffersink_pix_fmt(codec),
-        AV_PIX_FMT_NONE
-    };
-
-    err = av_opt_set_int_list(this->videoFilterSinkCtx, "pix_fmts",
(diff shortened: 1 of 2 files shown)


Madhav-arch:
Yeah, so, i tried downgrading, but ffmpeg is breaking other dependencies while doing so

ayuuuvauuu (approved):

joenano:
> Everyone interested, check #326, it has the patches we discussed in addition with the version checks so that it works with both the new and the old ffmpeg. I will merge #326 as soon as someone confirms that it works for them (i.e that I did not mess up the ffmpeg version numbers).

can confirm it works with latest arch version  2:8.0-3

PR #334, opened 2025-12-13
fix: wrong fish completion install path

### LINKED ISSUE #333: Incorrect Fish completion location
https://github.com/ammen99/wf-recorder/blob/v0.6.0/meson.build#L131 has the location for Fish completion as `share/fish/fish/vendor_completions.d/` when it should instead be `share/fish/vendor_completions.d/`

Thanks!

unlimitedsola:
Fixes #333


ammen99 (approved):
lgtm, thanks!

PR #339, opened 2026-02-07
Readme changes : Put each distro’s installation in a table ; add deps install for void

AKArien:
Thought it was neater like that :)


ammen99 (approved):
looks undoubtedly better :)

PR #348, opened 2026-06-25
Add more examples on a dedicated documentation page

spacebanana420:
This pull request adds a new file in `doc/examples.md` that contains a more extensive list of examples on how to record using wf-recorder, adding a deeper insight into FFmpeg capabilities.

It mentions lossless and RGB recording, intermediate formats and inter-frame compression, scaling the video, setting up x264 and more.

I also changed the readme and added a few small headers to the examples that were already in there. I kept the original examples in the readme, but I think we could also move them to the examples page or at least part of them.

I kept the manpage the same to not clutter it, similar to how the readme wasn't cluttered.


spacebanana420:
I noticed that this repo has a wiki and it has doumentation+examples. This PR would also suit well that wiki.

